PR #2, opened 2020-07-08
Feature/wketchum transfer existing

wesketchum:
Taking a stab at this for SBNSoftware/sbncode#9. 

Have added in ICARUS PMT Trigger data objects, and CRT data objects across the board. For the CRT ones, I've made 'common' ones for CRTHit, CRTTrack, and CRTTzero, though I've also copied the existing versions in each namespace because I worried about backwards compatibility (those are in the `_Legacy.hh` file). We should consider if we need that/when we can drop it.

I've just allowed experiement-specific CRTData objects (with only namespaces differing to distinguish the names). We should also consider if we can merge those or what.


cjbackhouse:
Thanks!

Do you have branches doing the removal from sbndcode and icaruscode?

PetrilloAtWork:
I am working on that now.

PR #9, opened 2021-02-02
Revert new products to use TVector's instead of geo types.

gputnam:
During the porting to the new repository structure, a bunch of TVector's were apparently changed to geo-types. This breaks a bunch of existing code (including code that was merged into sbncode).

This commit reverts those changes.


wesketchum:
OK, agree that this is fine for now since this won't be used in tons of data products. May need to revisit in the future, but better to not break code. Sorry for that.

PR #10, opened 2021-02-24
ICARUS PMT readout configuration data products.

PetrilloAtWork:
I am adding data products that hold PMT readout configuration.
ICARUS has a producer module that can extract this information from the FHiCL configuration set by DAQ (good at decoding time) and put in this easily accessible data structure.


wesketchum:
@PetrilloAtWork given the V1730s are used in both SBND and ICARUS, is there any reason to think this is strictly "experiment-specific"? Could move the location of the data object here now, and consider the location of that producer module too.

PetrilloAtWork:
Thank you for the question.
I think that the data product can indeed be shared between the experiments given that the boards are the same.
Note that the content of these data product(s) is currently quite minimal — I except it to be expanded as we go.
About the module used to extract the information: that is strictly dependent on the format of the FHiCL configuration used for DAQ.
I have seen it being already quite garbled (e.g. `board_id`, set, and `boardId`, unset, in the same configuration, baseline channel index being 1-based and threshold being 0-based, and so on) and I don't have a good perception of how that translates into SBND.
Ah, also the off-line channel mapping feature depends on a service that is implemented in ICARUS (but it does have an abstract interface so it could in principle be independently implemented for SBND too).

I let you take a decision for me: do you prefer me to move the thing to `Common`? and still `Common/PMT/Data`?

wesketchum:
I think we'll be working on some cleanup of the PMT configurations soon, so hopefully that will help with the module.

I'd say let's move it to `Common/PMT/Data` for the object. We can perhaps for the moment leave the module in `icaruscode` and imagine if it can be made more standard in the future...

Does that sound ok?

PetrilloAtWork:
The relevant data products in the branch have been moved from `sbnobj/ICARUS/PMT/Data` to `sbnobj/Common/PMT/Data` directory, and from `icarus` namespace into `sbn` (I thought it would be faster... :-P).
Modules filling these data products still live in `icaruscode` and, for the time being, they are not going to be moved.

PR #11, opened 2021-03-22
SBNEventWeight

mastbaum:
This pull request adds data products for SBN event weights (stored on the `art::Event`s) and parameters (stored on the `art::Run`s). The classes are located in a new directory `Common/SBNEventWeight`.

The former (weights) are simply a `typedef`, while the latter (parameters) are a new class. That class stores data in public members using only basic and STL types. Member functions related to initialization do have nontrivial logic and have library (CLHEP) dependencies, and currently handle error conditions with `assert` to avoid any LArSoft (specifically `cetlib`) dependence. The initialization code can be relocated to e.g. `sbncode` if preferred.

This should be reviewed alongside the `sbndcode` Pull Request which contains the framework and calculator code: https://github.com/SBNSoftware/sbncode/pull/97

### DIFF
--- a/sbnobj/Common/CMakeLists.txt
+++ b/sbnobj/Common/CMakeLists.txt
@@ -1,6 +1,6 @@
+add_subdirectory(Analysis)
 add_subdirectory(CRT)
 add_subdirectory(PMT)
 add_subdirectory(Reco)
-add_subdirectory(Analysis)
-
+add_subdirectory(SBNEventWeight)
 
--- a/sbnobj/Common/SBNEventWeight/CMakeLists.txt
+++ b/sbnobj/Common/SBNEventWeight/CMakeLists.txt
@@ -1,0 +1,16 @@
+cet_make( 
+  NO_DICTIONARY
+  LIBRARIES
+    cetlib_except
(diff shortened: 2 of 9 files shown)


sbnobj/Common/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
+add_subdirectory(Analysis)
```
cjbackhouse, line 1: Is the inclusion of this line an error? I guess it's personal code that you don't want merged?

mastbaum: That's the existing [`sbnobj/Common/Analysis` directory](https://github.com/SBNSoftware/sbnobj/tree/develop/sbnobj/Common/Analysis) — just moved up from line 3 to alphabetize for readability.

cjbackhouse: #headdesk

wesketchum:
Somehow this didn't get fully merged into v09_11_03 (sbncode v09_18_00_02).

Merging to dev now, and this last largely cosmetic commit will get into the next release.

PR #13, opened 2021-04-23
Updates to Stub objects

gputnam:
A few updates to the objects saved by stub reconstruction.

### DIFF
--- a/sbnobj/Common/Reco/Stub.cxx
+++ b/sbnobj/Common/Reco/Stub.cxx
@@ -1,5 +1,27 @@
 #include "sbnobj/Common/Reco/Stub.h"
 
+int sbn::Stub::PlaneIndex(const geo::PlaneID &p) const {
+  for (unsigned i_p = 0; i_p < plane.size(); i_p++) {
+    if (plane[i_p] == p) return i_p;
+  }
+
+  return -1;
+}
+
+float sbn::Stub::CoreCharge(const geo::PlaneID &p) const {
+  int plane_index = PlaneIndex(p);
+  if (plane_index < 0) return 0.;
+
+  return CoreCharge(plane_index);
+}
+
+int sbn::Stub::CoreNHit(const geo::PlaneID &p) const {
+  int plane_index = PlaneIndex(p);
+  if (plane_index < 0) return 0;
+
+  return CoreNHit(plane_index);
+}
+
 float sbn::Stub::CoreCharge(unsigned plane_index) const {
   if (plane_index >= plane.size()) return -1;
   float ret = 0.;
@@ -33,14 +55,12 @@ int sbn::Stub::CoreNHit(unsigned plane_index) const {
 }
     
 bool sbn::Stub::OnCore(const geo::WireID &w) const {
-  geo::PlaneID p = w;
-  for (unsigned plane_index = 0; plane_index < plane.size(); plane_index++) {
-    if (plane[plane_index] != p) continue;
+  int plane_index = PlaneIndex(w);
+  if (plane_index < 0) return -1;
 
-    int stubdir = vtx_w[plane_index] <= hit_w[plane_index] ? 1 : -1;
-    bool before_vtx = (((int)w.Wire - vtx_w[plane_index]) * stubdir) < 0;
-    bool after_hit = (((int)w.Wire - hit_w[plane_index]) * stubdir) > 0;
-    return !before_vtx && !after_hit;
-  }
-  return false;
+  int stubdir = vtx_w[plane_index] <= hit_w[plane_index] ? 1 : -1;
+  bool before_vtx = (((int)w.Wire - vtx_w[plane_index]) * stubdir) < 0;
+  bool after_hit = (((int)w.Wire - hit_w[plane_index]) * stubdir) > 0;
+
+  return !before_vtx && !after_hit;
 }
--- a/sbnobj/Common/Reco/Stub.cxx
+++ b/sbnobj/Common/Reco/Stub.cxx
@@ -1,0 +1,34 @@
+#include "sbnobj/Common/Reco/Stub.h"
+
+float sbn::Stub::CoreCharge(unsigned plane_index) const {
+  if (plane_index >= plane.size()) return -1;
+  float ret = 0.;
+
+  int stubdir = vtx_w[plane_index] <= hit_w[plane_index] ? 1 : -1;
+  for (const sbn::StubHit &h: hits[plane_index]) {
+    bool before_vtx = ((h.wire - vtx_w[plane_index]) * stubdir) < 0;
+    bool after_hit = ((h.wire - hit_w[plane_index]) * stubdir) > 0;
+    if (!before_vtx && !after_hit) {
+      ret += h.charge;
+    }
+  }
+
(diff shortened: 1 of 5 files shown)


sbnobj/Common/Reco/Stub.cxx
```diff
@@ -0,0 +1,46 @@
...
+  for (unsigned plane_index = 0; plane_index < plane.size(); plane_index++) {
+    if (plane[plane_index] != p) continue;
+
+    int stubdir = vtx_w[plane_index] <= hit_w[plane_index] ? 1 : -1;
+    bool before_vtx = (((int)w.Wire - vtx_w[plane_index]) * stubdir) < 0;
+    bool after_hit = (((int)w.Wire - hit_w[plane_index]) * stubdir) > 0;
+    return !before_vtx && !after_hit;
+  }
+  return false;
+}
```
PetrilloAtWork, line 46: Just a style comment here. If I understand it well, the actual content of the loop here is supposed to be executed only once. Hence, it does not really belong to the loop. Separating it would make it easier to read the code.
Example:
~~~~~~.cpp
bool sbn::Stub::OnCore(const geo::WireID &w) const {
  
  // find the plane
  geo::PlaneID const& p = w;
  unsigned plane_index = 0U;
  while (plane_index < plane.size()) {
    if (plane[plane_index] == p) break;
    ++plane_index;
  }
  if (plane_index == plane.size()) return false;

  // core check
  int stubdir = vtx_w[plane_index] <= hit_w[plane_index] ? 1 : -1;
  bool before_vtx = (((int)w.Wire - vtx_w[plane_index]) * stubdir) < 0;
  bool after_hit = (((int)w.Wire - hit_w[plane_index]) * stubdir) > 0;
  return !before_vtx && !after_hit;
}
~~~~~~

sbnobj/Common/Reco/Stub.h
```diff
@@ -6,16 +6,33 @@
...
+    std::vector<float> pitch; //!< Pitch of stub on each wire [cm]
+    std::vector<float> trkpitch; //!< Pitch of the matched track on each wire [cm]
+    std::vector<float> vtx_w; //!< Wire coordinate of the vertex on this plane
+    std::vector<short> hit_w; //!< Wire of the end point hit on this plane
+
+    std::vector<std::vector<StubHit>> hits; //!< Hits on each plane. Ordered vtx->end
+
+    /// Helper functions
+    float CoreCharge(unsigned plane_index=0) const; //!< Returns the charge along the core of the stub on the given plane index
+    int CoreNHit(unsigned plane_index=0) const; //!< Returns the number of hits along the core of the stub on the given plane index
```
wesketchum, line 33: Is there an interface to giving a geo::PlaneID and getting this information? I'm worried this may be a little confusing to have to go through the list of 'planes' which are sorted by quality rather than number. Think this is likely to be misinterpreted...

sbnobj/Common/Reco/classes_def.xml
```diff
@@ -72,14 +72,24 @@
   <class name="art::Wrapper<art::Assns<float,recob::Shower,void>>" />
   <class name="art::Wrapper<art::Assns<recob::Shower,float,void>>" />
 
-  <class name="sbn::Stub" ClassVersion="11">
+  <class name="sbn::Stub" ClassVersion="16">
+   <version ClassVersion="16" checksum="1338933861"/>
+   <version ClassVersion="15" checksum="961565117"/>
+   <version ClassVersion="14" checksum="2410976659"/>
+   <version ClassVersion="13" checksum="2883383711"/>
+   <version ClassVersion="12" checksum="703271609"/>
```
wesketchum, line 80: Are the versions 12-16 there due to changes you made during development? If so, probably better to just declare this update version 12 and remove the rest?

PetrilloAtWork (commented):
I have thrown some comments. Let me know what do you think...

PetrilloAtWork (commented):
... and I have forgotten a general comment and a half.

First, `sbn::Stub` contains several collections (`std::vector`) that are requested to have the same size, but no enforcement (nor check) is performed. That is the usual vector-of-struct vs. struct-of-vector fight. It is not fully clear to me which the constraints are, so I can't confidently say which approach is better here, but I would like you to think of that. In case, there are two ways to go toward the enforcement I was talking about; one is via interface (a `add()` method that requires all the values to be specified, optionally moving the actual data into `private`), and/or the aforementioned vector-of-struct, a single vector where each entry is an aggregate of all relevant information.

Second, my pet peeve: `TVector3` ~should be dropped, walked over and then burned~ is not well-designed according to modern paradigms (in short: it derives from `TObject`).
Support in ROOT for the newer `GenVector` vectors has been largely improved. The key aliases are `ROOT::Math::XYZPoint` and `ROOT::Math::XYZVector`. I am a persevering proponent of moving to them.

sbnobj/Common/Reco/Stub.cxx
gputnam: Fixed.

sbnobj/Common/Reco/Stub.h
gputnam: Fixed.

sbnobj/Common/Reco/classes_def.xml
gputnam: Fixed.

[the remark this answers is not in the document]
gputnam: Fixed.

sbnobj/Common/Reco/VertexHit.h
```diff
@@ -11,6 +11,7 @@ namespace sbn {
     geo::WireID wire;    
     float charge;
     float proj_dist_to_vertex;
+    float vtxw;
```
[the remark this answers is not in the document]
gputnam: Added documentation.

gputnam:
I have addressed the comments relayed here.

sbnobj/Common/Reco/VertexHit.h
```diff
@@ -8,18 +8,19 @@
...
+    float vtxw; //!< Wire of the vertex associated with this hit
+    int spID; //!< ID of the SpacePoint associated with this hit
+    TVector3 spXYZ; //!< 3D location of the SpacePoint associated with this hit
+    float pitch; //!< Computed pitch of a track traversing from the vertex to this hit [cm]
+    float dqdx; //!< charge/pitch [#elec/cm]
+    float dedx; //!< Recombination corrected dQ/dx [MeV/cm]
+    std::vector<int> nearbyPFPIDs; //!< ID's of PFParticle's near this hit in 2D
+    std::vector<float> nearbyPFPDists; //!< 2D Dot product of vertex-hit direction to PFParticle direction
+    std::vector<float> nearbyPFP3DDists; //!< ID's of PFPArticle's near this hit in 3D
+    std::vector<int> nearbyPFP3DIDs; //!< 3D Dot product of vertex-hit direction to PFParticle direction
```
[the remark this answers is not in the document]
PetrilloAtWork: Gray wrote me privately that the check has been executed and passed.

PetrilloAtWork (approved):
Good to go!

PR #15, opened 2021-05-19
Feature/jzettleNewFMObject

jzettle:
Pull request for the addition of a new sbn::SimpleFlashMatch data product replacing the anab::T0 object that was used previously. This is largely done to improve the amount of stored information from the simple flash match object in the CAF files. Tested on SBND and ICARUS simulations and shown to fill the correct object in the CAF files.

### DIFF
--- a/sbnobj/Common/Reco/SimpleFlashMatchVars.h
+++ b/sbnobj/Common/Reco/SimpleFlashMatchVars.h
@@ -9,14 +9,16 @@ namespace sbn
   {
   public:
   SimpleFlashMatch(
-    bool present = false, double time = -1, double pe = -1,
+    bool present = false, double time = -1,
+    double charge_q = -1, double light_pe = -1,
     double score = -1, double scr_y = -1, double scr_z = -1,
     double scr_rr = -1, double scr_ratio = -1,
     TVector3 chargeXYZ = TVector3(-999, -999, -999),
     TVector3 lightXYZ = TVector3(-999,-999,-999)):
     mPresent(present),
     mTime(time),
-    mPE(pe),
+    mChargeQ(charge_q),
+    mLightPE(light_pe),
     mScore(score),
     mScr_y(scr_y),
     mScr_z(scr_z),
@@ -28,7 +30,8 @@ namespace sbn
 
     bool mPresent;
     double mTime;
-    double mPE;
+    double mChargeQ;
+    double mLightPE;
     double mScore;
     double mScr_y;
     double mScr_z;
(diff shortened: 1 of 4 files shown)


ikatza:
@FernandaPsihas could you please review this?

We've tested to be able to produce car and flatcafs, but let us know if there are some styling or conventions that we might have missed.

sbnobj/Common/Reco/SimpleFlashMatchVars.h
```diff
@@ -0,0 +1,45 @@
...
+    bool mPresent;
+    double mTime;
+    double mChargeQ;
+    double mLightPE;
+    double mScore;
+    double mScr_y;
+    double mScr_z;
+    double mScr_rr;
+    double mScr_ratio;
+    TVector3 mChargeXYZ;
```
cjbackhouse, line 40: I know lots of people have Opinions about TVector3, not without justification. There are larsoft 3-vector types that are supposed to be better. Can you use one of those?

sbnobj/Common/Reco/SimpleFlashMatchVars.h
```diff
@@ -0,0 +1,45 @@
...
+    mScore(score),
+    mScr_y(scr_y),
+    mScr_z(scr_z),
+    mScr_rr(scr_rr),
+    mScr_ratio(scr_ratio),
+    mChargeXYZ(chargeXYZ),
+    mLightXYZ(lightXYZ)
+    {}
+
+    bool mPresent;
```
cjbackhouse, line 31: The convention elsewhere in larsoft is `f` for field, rather than `m` for member. Unless this is matching the style of some related existing code?

sbnobj/Common/Reco/SimpleFlashMatchVars.h
ikatza: Good point.
How about using `geo::Point_t` from `#include "larcoreobj/SimpleTypesAndConstants/geo_vectors.h" `?

sbnobj/Common/Reco/SimpleFlashMatchVars.h
jzettle: This was the convention in place in the old object when the edits were made although we're happy to change to a different convention. Personally I'm used to `f` in this instance as you mention as well.

sbnobj/Common/Reco/SimpleFlashMatchVars.h
ikatza: Come to think about it. Whilst I'm all for removing `Tvector3`, there are several other instances of it on `sbnobj`. 
Better to make one sweep change in another PR.

ikatza:
Unless self reviewing is acceptable, makes no sense for me to review this PR.

sbnobj/Common/Reco/SimpleFlashMatchVars.h
ikatza: Looking at other `sbnobj` classes I see that it's fairly common to not have `m` or `f`. So I decided to remove clutter.

FernandaPsihas (approved):

PR #16, opened 2021-05-21
Added TimeTagForTimeStamp parameter to PMT configuration data object

PetrilloAtWork:
This is actually DAQ configuration rather than PMT readout.
I'd say "close enough".


wesketchum (approved):
Looks fine!

PR #17, opened 2021-06-17
New track calo skimmer module for ICARUS calibration stuff.

gputnam:
New objects. Orthogonal to existing stuff.

### DIFF
--- a/sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
+++ b/sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
@@ -129,6 +129,7 @@ namespace sbn {
     int n_fit_point;
 
     int selected;
+    int nprescale;
 
     std::vector<float> tracks_near_end_dist;
     std::vector<float> tracks_near_end_costh;
@@ -167,7 +168,8 @@ namespace sbn {
       exp_fit_R(-1),
       exp_fit_residuals(-1),
       n_fit_point(-1),
-      selected(-1) {}
+      selected(-1),
+      nprescale(-1) {}
   };
 
 }
--- a/sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
+++ b/sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
@@ -1,0 +1,156 @@
+#ifndef SBN_TrackCaloSkimmerObj
+#define SBN_TrackCaloSkimmerObj
+
+#include <vector>
+
+namespace sbn {
+  struct HitInfo {
+    float integral;
+    float sumadc;
+    float width;
+    float pitch;
+    float dqdx;
+    float rr;
+    float x;
+    float y;
+    float z;
+    float time;
+    uint16_t wire;
+    uint16_t plane;
+    uint16_t tpc;
+    uint16_t mult;
+    uint16_t i_snippet;
+    int16_t start;
+    int16_t end;
+    bool ontraj;
+    bool oncalo;
+
+    // raw adc info
+    short tdc0;
+    std::vector<short> adcs;
+
+    HitInfo():
+      integral(-1),
+      sumadc(-1),
+      width(-1),
+      pitch(-1),
+      dqdx(-1),
+      rr(-1),
+      x(-1),
+      y(-1),
+      z(-1),
+      time(-1),
+      wire((uint16_t)-1),
+      plane((uint16_t)-1),
+      tpc((uint16_t)-1),
+      mult((uint16_t)-1),
+      i_snippet((uint16_t)-1),
+      start(-1),
+      end(-1),
+      ontraj(false),
+      oncalo(false),
+      tdc0(-1) {}
+  };
+
+  struct MetaInfo {
+    int run;
+    int evt;
+    int subrun;
+    int ifile;
+    int iproc;
+
+   MetaInfo():
+     run(-1),
+     evt(-1),
+     subrun(-1),
+     ifile(-1),
+     iproc(-1) {}
+  };
+
+  struct TrackInfo {
+    MetaInfo meta;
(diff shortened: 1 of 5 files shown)


sbnobj/Common/Calibration/classes_def.xml
```diff
@@ -0,0 +1,6 @@
+<lcgdict>
+  <class name="sbn::TrackInfo" />
```
wesketchum, line 2: Include ClassVersion here (and in the classes below). Will allow for version updating/reading old products if these persist for whatever reason.

sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
```diff
@@ -0,0 +1,156 @@
...
+     run(-1),
+     evt(-1),
+     subrun(-1),
+     ifile(-1),
+     iproc(-1) {}
+  };
+
+  struct TrackInfo {
+    MetaInfo meta;
+    std::vector<HitInfo> hits0;
```
wesketchum, line 72: Is there a reason to include the hitinfo in the TrackInfo object like this? Rather than do an association (in art)? Or is this for ease/convenience of setting up the output trees? I think we'd generally prefer to do a linkage by association rather than inclusion like this.

wesketchum (commented):
Generally looks ok. Few comments below.

sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
```diff
@@ -0,0 +1,156 @@
...
+    float end_y;
+    float end_z;
+    float dir_x; 
+    float dir_y;
+    float dir_z;
+    float length;
+
+    unsigned ndaughters;
+
+    float hit_min_time_p0_tpcE;
```
wesketchum, line 91: Actually, a little confused here. this is per plane, but then also ... per cryostat? there are two physical TPCs in east and west, cryostats, so not sure what these hits correspond to?

Might suggest that this be keyed by PlaneID (e.g. a map of floats?) to make it a bit more general to SBND too. Think ROOT should handle maps ok.

gputnam: It's per-plane per-TPC (E/W). I believe that structure should also generalize to SBND.

sbnobj/Common/Calibration/classes_def.xml
gputnam: Gotcha -- how do you do that exactly? I am not familiar.

sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
gputnam: This structure isn't really intended for use in art. It's just supposed to represent a relatively flat structure that can be encoded as an ntuple. So I don't think bringing in art things makes sense here.

sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
wesketchum: Sorry, maybe I'm misunderstanding: is this already per cryostat somehow? So, one would do this for the west cryostat, and then again for the easy cryostat, and those would be two separate collections?

sbnobj/Common/Calibration/classes_def.xml
wesketchum: See here: https://github.com/SBNSoftware/sbnobj/blob/867ce19432565d983144faa2e0ff41cec59d09a6/sbnobj/Common/Reco/classes_def.xml#L2

You just need to add in the ClassVersion part, and when you compile it should automatically calculate and add the checksum.

sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
gputnam: Each TrackInfo object corresponds to a single reconstructed track, and each track is always in one cryostat. So it is "per-cryostat" in that sense.

sbnobj/Common/Calibration/classes_def.xml
gputnam: Fixed.

gputnam:
Added in ClassVersion's/

sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
wesketchum: OK, I still suggest this be keyed per PlaneID, as that will be general and is what you want. It may be very much useful to know which cryostat the tracks you're looking at come from (e.g. purity is different), and so I would not lose that information.

gputnam: I want to be able to access this data outside of the LArSoft context, which is why I have avoided using any LArSoft-specific objects. But I'll add in a cryostat field.

gputnam: Fixed.

wesketchum: OK, One last thing is that I noticed the cryostat is not included in the HitInfo, which I think you want to add.

gputnam: hmm... I don't think that's necessary.

Right now the HitInfo is the most common object, so I want to keep its size as small as possible. And tracks are by definition limited to a cryostat, so the hits are all in the same cryostat as the track. So one can obtain the hit Cryostat from the TracjInfo.

wesketchum: Fair enough.

wesketchum (approved):

PR #18, opened 2021-06-17
Feature/zennamo bnb spill accounting

jzennamo:
This is needed to store beam spill information in our files, this is fairly compact ~2.7 kb/event and 0.07 kb/spill

### DIFF
--- a/sbnobj/Common/POTAccounting/BNBSpillInfo.h
+++ b/sbnobj/Common/POTAccounting/BNBSpillInfo.h
@@ -1,10 +1,10 @@
-#ifndef sbncode_BeamSpillInfo_HH
-#define sbncode_BeamSpillInfo_HH
+#ifndef sbncode_BNBSpillInfo_HH
+#define sbncode_BNBSpillInfo_HH
 
 #include <vector>
 
 namespace sbn {
-  class BeamSpillInfo {
+  class BNBSpillInfo {
   public:
     //typedef enum triggertype { kBNB, kEXT, kUnbiased, kNUMI } TriggerType; //!< Which beamline are we looking at
     
--- a/sbnobj/Common/POTAccounting/CMakeLists.txt
+++ b/sbnobj/Common/POTAccounting/CMakeLists.txt
@@ -1,0 +1,11 @@
+cet_make( 
+  LIBRARIES
+    cetlib_except
+    ${ROOT_BASIC_LIB_LIST}
+  NO_DICTIONARY
+  )
+
+art_dictionary(DICTIONARY_LIBRARIES sbnobj_Common_POTAccounting)
+
+install_headers()
+install_source()
--- a/sbnobj/Common/POTAccounting/classes.h
+++ b/sbnobj/Common/POTAccounting/classes.h
@@ -1,0 +1,9 @@
+#include <map>
+#include <string>
+#include <vector>
+#include "canvas/Persistency/Common/Wrapper.h"
+#include "canvas/Persistency/Common/Assns.h"
+#include "sbnobj/Common/POTAccounting/BeamSpillInfo.h"
+
+
+
--- a/sbnobj/Common/POTAccounting/classes.h
+++ b/sbnobj/Common/POTAccounting/classes.h
@@ -3,7 +3,7 @@
 #include <vector>
 #include "canvas/Persistency/Common/Wrapper.h"
 #include "canvas/Persistency/Common/Assns.h"
-#include "sbnobj/Common/POTAccounting/BeamSpillInfo.h"
+#include "sbnobj/Common/POTAccounting/BNBSpillInfo.h"
 
 
 
--- a/sbnobj/Common/POTAccounting/classes_def.xml
+++ b/sbnobj/Common/POTAccounting/classes_def.xml
@@ -1,13 +1,14 @@
 <lcgdict>
-  <class name="sbn::BeamSpillInfo" ClassVersion="14">
+  <class name="sbn::BNBSpillInfo" ClassVersion="15">
+   <version ClassVersion="15" checksum="1226979720"/>
    <version ClassVersion="14" checksum="2089752727"/>
    <version ClassVersion="13" checksum="1538708656"/>
    <version ClassVersion="12" checksum="2483311428"/>
    <version ClassVersion="11" checksum="4133298356"/>
    <version ClassVersion="10" checksum="13541"/>
   </class>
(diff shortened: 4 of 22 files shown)


sbnobj/Common/POTAccounting/BNBSpillInfo.h
```diff
@@ -0,0 +1,63 @@
...
+
+#include <vector>
+
+namespace sbn {
+  class BNBSpillInfo {
+  public:
+    //typedef enum triggertype { kBNB, kEXT, kUnbiased, kNUMI } TriggerType; //!< Which beamline are we looking at
+    
+    unsigned long int spill_time_s; //!< The IFDB Beam Spill Time, unit sec
+    unsigned long int spill_time_ns; //!< The IFDB Beam Spill Time, unit nsec
```
PetrilloAtWork, line 12: Is there a spill ID in the database? if there is any time of single key column in that DB, it may be useful to store it here for effective lookup.

sbnobj/Common/POTAccounting/BNBSpillInfo.h
```diff
@@ -0,0 +1,63 @@
...
+
+    double HPTG1; //!< Horizontal Position Monitor at Target Station 1, units mm 
+    double VPTG1; //!< Horizontal Position Monitor at Target Station 1, units mm
+
+    double HPTG2; //!< Horizontal Position Monitor at Target Station 2, closest to target, units mm
+    double VPTG2; //!< Horizontal Position Monitor at Target Station 2, closest to target, units mm
+    
+    double BTJT2; //!< Temperature of air exiting target, units Deg C
+
+    double THCURR; //!< Current applied to Horn, units kiloAmperes
```
PetrilloAtWork, line 37: I actually doubt that any of these measures would not be fitting in a single precision number (6-7 significant digits).
Unless there is some large offset, that is (like measuring the position of the beam with the location of the Director's office as reference).
Moving to single precision would save quite some memory, and I _suggest_ this to be done if feasible.

sbnobj/Common/POTAccounting/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+cet_make( 
+  LIBRARIES
+    cetlib_except
+    ${ROOT_BASIC_LIB_LIST}
+  NO_DICTIONARY
+  )
+
```
PetrilloAtWork, line 7: I don't think this is necessary at all.

sbnobj/Common/POTAccounting/BNBSpillInfo.h
```diff
@@ -0,0 +1,63 @@
+#ifndef sbncode_BNBSpillInfo_HH
+#define sbncode_BNBSpillInfo_HH
```
PetrilloAtWork, line 2: Take ~responsibility~ ~ownership~ the glory!
```suggestion
/**
 * @file sbnobj/Common/POTAccounting/BNBSpillInfo.h
 * @author Joseph Zennamo (jaz8600@fnal.gov)
 */

#ifndef sbncode_BNBSpillInfo_H
#define sbncode_BNBSpillInfo_H
```
(also, the file name is `.h`, so the include guard should go `_H`)

sbnobj/Common/POTAccounting/classes.h
```diff
@@ -0,0 +1,9 @@
+#include <map>
+#include <string>
```
PetrilloAtWork, line 2: I don't think `<map>` and `<string>` are needed.

sbnobj/Common/POTAccounting/classes_def.xml
```diff
@@ -0,0 +1,14 @@
+<lcgdict>
+  <class name="sbn::BNBSpillInfo" ClassVersion="15">
+   <version ClassVersion="15" checksum="1226979720"/>
+   <version ClassVersion="14" checksum="2089752727"/>
+   <version ClassVersion="13" checksum="1538708656"/>
+   <version ClassVersion="12" checksum="2483311428"/>
+   <version ClassVersion="11" checksum="4133298356"/>
+   <version ClassVersion="10" checksum="13541"/>
```
PetrilloAtWork, line 8: Remove the development versions and make this be version `10`.

PetrilloAtWork (commented):
Some formal comments _to start_.
I may have more after seeing the other branch.

sbnobj/Common/POTAccounting/BNBSpillInfo.h
jzennamo: sadly there is not, that would make things MUCH easier. In fact, many devices actually report different times and we have to adjust those times or select the ones that are closest in time.

sbnobj/Common/POTAccounting/BNBSpillInfo.h
```diff
@@ -0,0 +1,63 @@
...
+    std::vector< double > VPTGT;
+    std::vector< double > HITGT;
+    std::vector< double > VITGT; 
+    std::vector< double > MTGTDS; //MULTIWIRES
+    double HRNDIR; // horn polarity (based on sign)
+    double NSLINA; // horn current /4 
+    double NSLINB; // horn current /4  
+    double NSLINC; // horn current /4 
+    double NSLIND; // horn current /4 
+    */
```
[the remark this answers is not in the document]
jzennamo: dropped these

sbnobj/Common/POTAccounting/BNBSpillInfo.h
jzennamo: switched to floats

sbnobj/Common/POTAccounting/BNBSpillInfo.h
jzennamo: switched

sbnobj/Common/POTAccounting/classes.h
jzennamo: dropped

sbnobj/Common/POTAccounting/classes_def.xml
jzennamo: I tried to do this but I think I failed...

jzennamo:
@PetrilloAtWork I added a set of variables that store the time offset between the matched multiwire chambers and the spill time, these will probably be needed to determine when to use the multiwires for beam quality cuts

jzennamo:
I have added a data-product to collect the number of EXT gates

cjbackhouse (approved):

PR #19, opened 2021-06-25
Razzle Dazzle PID

etyley:
Adding header files for MVA PID and associated data products, see sbncode PR of the same name for details.


sbnobj/Common/Reco/ScatterClosestApproach.h
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  ScatterClosestApproach(const float mean, const float stdDev, const float max)
+      : mMean(mean)
+      , mStdDev(stdDev)
+      , mMax(max)
+  {
+  }
+
+  float mMean;   // Average ClosestApproach [cm]
```
cjbackhouse, line 23: We use `f` for fields elsewhere. But I also think for a "struct-like" data product like this where the fields are directly accessible I just wouldn't prefix at all.

sbnobj/Common/Reco/MVAPID.cc
```diff
@@ -0,0 +1,25 @@
+#include "sbnobj/Common/Reco/MVAPID.h"
+
+#include "cetlib_except/exception.h"
+#include "messagefacility/MessageLogger/MessageLogger.h"
+
+#include <algorithm>
+
+void sbn::MVAPID::AddScore(int pdg, float score)
+{
+  auto [iter, success] = mMVAScoreMap.try_emplace(pdg, score);
```
cjbackhouse, line 10: I didn't know you could do multiple assignment like this in C++ yet, cool.

That said, you don't need the iter so I guess you could write `if(!mMVAScoreMap.try_emplace(pdg, score).second){`

sbnobj/Common/Reco/MVAPID.cc
```diff
@@ -0,0 +1,25 @@
...
+    throw cet::exception("MVAPID") << "Failed to add entry: " << pdg << " to MVA PID score map as it already exists" << std::endl;
+}
+
+std::pair<int, float> sbn::MVAPID::BestIter() const
+{
+  auto iter = std::max_element(mMVAScoreMap.begin(), mMVAScoreMap.end(), [](auto const& lhs, auto const& rhs) { return lhs.second < rhs.second; });
+
+  if (iter == mMVAScoreMap.end()) {
+    mf::LogError("MVAPID") << "Failed to find max element in map" << std::endl;
+    return std::make_pair<int, float>(-1, std::numeric_limits<float>::lowest());
```
cjbackhouse, line 21: I suspect you can just write `return {-1, std::numeric_limits<float>::lowest())};`?

sbnobj/Common/Reco/MVAPID.h
```diff
@@ -0,0 +1,24 @@
+// Class for holding the result of an MVA PID
+
+#ifndef sbncode_MVAPID_H
```
cjbackhouse, line 3: Capitalize include guards

sbnobj/Common/Reco/MVAPID.h
```diff
@@ -0,0 +1,24 @@
...
+namespace sbn {
+class MVAPID {
+  public:
+  MVAPID() {}
+
+  std::map<int, float> mMVAScoreMap; // MVA Scores: Key is PDG of hypothesis and value is the MVA score
+
+  void AddScore(int pdg, float score); // Function to call to add a new score to the map
+
+  std::pair<int, float> BestIter() const;                             // Funtion to return the map iterator with the highest score
```
cjbackhouse, line 18: Typo in comment. Also, we know they're functions. How about just "Return the map entry with the highest score"?

sbnobj/Common/Reco/MVAPID.cc
```diff
@@ -0,0 +1,25 @@
+#include "sbnobj/Common/Reco/MVAPID.h"
```
cjbackhouse, line 1: I was going to comment that implementation files should be .cxx and that .cc is for modules. But I see we have both conventions within sbnobj already :sob:

sbnobj/Common/Reco/MVAPID.h
```diff
@@ -0,0 +1,24 @@
+// Class for holding the result of an MVA PID
+
+#ifndef sbncode_MVAPID_H
+#define sbncode_MVAPID_H
+
+#include <map>
+#include <string>
```
cjbackhouse, line 7: No need for string here

sbnobj/Common/Reco/StoppingChi2Fit.h
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  StoppingChi2Fit(const float Pol0Chi2, const float ExpChi2, const float Pol0Fit)
+      : mPol0Chi2(Pol0Chi2)
+      , mExpChi2(ExpChi2)
+      , mPol0Fit(Pol0Fit)
+  {
+  }
+
+  float mPol0Chi2; // Chi^2 of fitting a 0 degree polynomial
```
cjbackhouse, line 23: Same comment about removing the `m`s

cjbackhouse (changes requested):
Fine, but I'm requesting variable name changes before they get baked into files.

sbnobj/Common/Reco/MVAPID.h
etyley: I was just following what has been done in other files: 
```FlashTriggerPrimitive.hh:#ifndef FlashTriggerPrimitive_SBN_hh_
MergedTrackInfo.hh:#ifndef sbncode_MergedTrack_HH
MVAPID.h:#ifndef sbncode_MVAPID_H
RangeP.h:#ifndef sbncode_rangep_HH
ScatterClosestApproach.h:#ifndef sbncode_scatterclosestapproach_H
ShowerSelectionVars.h:#ifndef sbncode_Showerselectionvars_H
SimpleFlashMatchVars.h:#ifndef sbncode_Simpleflashmatchvars_H
StoppingChi2Fit.h:#ifndef sbncode_stoppingchi2fit_H
Stub.h:#ifndef sbncode_Stub_HH
VertexHit.h:#ifndef sbncode_VertexHit_HH
```

I am happy to update either mine or all of them if this is something you feel strongly on?

cjbackhouse: Maybe let's make another PR that changes all the include guards and renames the cc's to cxx's? The problem will come when it turns out that someone has been enforcing the opposite convention to me...

etyley: Okay, that sounds good to me. I assume you have a strong opinion on `.h` vs `.hh` which are also used interchangeably currently and should be fixed in that PR

cjbackhouse: `.h` please. Although changing those are presumably going to have an impact across all sorts of downstream repositories.

etyley: Yeah, that will also require PRs in sbncode (and maybe sbndcode and icaruscode). If it sounds good to you, I will wait until this PR is merged then go through and update this?

etyley:
I think I have addressed all of your comments

sbnobj/Common/Reco/ScatterClosestApproach.h
cjbackhouse: sorry, can we lowercase the initial letters too? (I assume `max` doesn't cause trouble by clashing with a function?)

sbnobj/Common/Reco/MVAPID.h
cjbackhouse: :+1:

sbnobj/Common/Reco/StoppingChi2Fit.h
cjbackhouse: same here

sbnobj/Common/Reco/ScatterClosestApproach.h
etyley: Done

cjbackhouse (approved):

PR #20, opened 2021-07-10
Refactor configs and products out of sbncode for MeVPrtl Event Generator

gputnam:
Adds in products to sbnobj for MeVPrtl Event Generation.


mroda88 (approved):
All looks good

PR #22, opened 2021-07-20
Updates to TrackSkim object. Add in end-point-hit information and truth information.

gputnam:
Updates to objects for track-calo skimming in ICARUS. Adds in truth information and hits near the end-point of the track.


mrmooney (approved):
Looks good to go

PR #23, opened 2021-07-25
Fix to Missing FileEntry for sbnobj/Common/Calibration/TrackCaloSkimmerObj.h

PetrilloAtWork:
_Edit: moved the explanation into an issue ticket (#24) for ease of lookup._

The header in this directory, `sbnobj/Common/Calibration/TrackCaloSkimmerObj.h`, was not installed.
This prevents on-the-fly compilation by ROOT, and therefore loading files with this data object.

While I have tested the fix itself, **this is a GitHub commit, which has not been tested _at all_.** Integration test should tell.

### DIFF
--- a/sbnobj/Common/Calibration/CMakeLists.txt
+++ b/sbnobj/Common/Calibration/CMakeLists.txt
@@ -1,1 +1,4 @@
 art_dictionary()
+
+install_source()
+install_headers()


PetrilloAtWork:
I also bless you with my opinion that given the type of change this fix may end up in `v09_11_15`, that is rewriting the history for `master` or adding the missing header manually to the UPS package.

gputnam (approved):
Looks good! Thanks Gianluca — my bad for missing this.

SFBayLaser (approved):
This looks good to me!

PR #28, opened 2021-10-07
add wdith2 for pmNsigma mode

Ooohu:
Update on `sbnobj/Common/SBNEventWeight/EventWeightParameterSet.*` to match up the PR #176 at sbncode: https://github.com/SBNSoftware/sbncode/pull/176

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,9 +13,9 @@
 # ======================================================================
 
 
-cmake_minimum_required (VERSION 3.12...3.18 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.19 FATAL_ERROR)
 
-project(sbnobj LANGUAGES CXX)
+project(sbnobj VERSION 09.12.05 LANGUAGES CXX)
 
 message(STATUS
   "\n-- ============================================================================="
@@ -46,6 +46,7 @@ find_ups_product( lardataobj )
 find_ups_product( larcorealg )
 find_ups_product( lardataalg )
 find_ups_product( nusimdata )
+find_ups_product(clhep)
 find_ups_boost( )
 find_ups_root()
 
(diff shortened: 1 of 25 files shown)


mroda88 (commented):
@Ooohu I'm sorry but this PR does not make any sense and we never discussed it before, so you need to explain a few things. 

What is the meaning of the second width? What is the usage case?
Your descriptions says
```
Gaussian sigma, if one more variation is needed
```
Which is not very descriptive. 
Also this does not provide a description of the changes for the fcl files.

Ooohu:
> @Ooohu I'm sorry but this PR does not make any sense and we never discussed it before, so you need to explain a few things.
> 
> What is the meaning of the second width? What is the usage case? Your descriptions says
> 
> ```
> Gaussian sigma, if one more variation is needed
> ```
> 
> Which is not very descriptive. Also this does not provide a description of the changes for the fcl files.

Hello @mroda88, this description is for the lines 
https://github.com/SBNSoftware/sbnobj/blob/0ce442f288b08a99b52045bad04446367e7b7bc8/sbnobj/Common/SBNEventWeight/EventWeightParameterSet.cxx#L71:

```     
else if (fRWType == kPMNSigma) {
        it.second.push_back(p.fMean + p.fWidth);
        it.second.push_back(p.fMean + p.fWidth2);
      }
      ```
     
     I will update the description from
     ` float fWidth2;  //!< Gaussian sigma, if one more variation is needed` 
     to
     `  float fWidth2;  //!< for PMNSigma mode, which will requires two shifts to the knob: fmean+fWidth1 and fmean + fWidth2`.
     
When using PMNSigma mode, two universes (with two shifted values) are generated, and this is the reason why `fWidth2` is used.

Ooohu:
Now only change the line at https://github.com/SBNSoftware/sbnobj/blob/337599381d1786c98aaf1e0d0734819a449643ce/sbnobj/Common/SBNEventWeight/EventWeightParameterSet.cxx#L64

mroda88 (approved):
great, thanks

PR #29, opened 2021-10-14
Save More Truth Information in Calibration NTuple

gputnam:
Also save channel number for per-wire stuff

### DIFF
--- a/sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
+++ b/sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
@@ -69,7 +69,7 @@ namespace sbn {
     HitInfo h; //!< Hit information by itself
     float pitch; //!< Pitch of track across wire the hit is on [cm]
     float dqdx; //!< Initial computed dq/dx of hit [ADC/cm]
-    float rr; //!< Residual range of hit along track [sm]
+    float rr; //!< Residual range of hit along track [cm]
     Vector3D dir; //!< Direction of track at hit location
     uint16_t i_snippet; //!< Index of hit into snippet 
     bool ontraj; //!< Whether the hit is on the track trajectory
@@ -105,16 +105,16 @@ namespace sbn {
     int16_t cryo; //!< Cryostat of hit
     int16_t tpc; //!< TPC of hit
     int16_t plane; //!< Plane of hit
-    int wire; //!< Wire of hit
-    int channel; //!< Channel of hit
+    int16_t wire; //!< Wire of hit
+    int16_t channel; //!< Channel of hit
 
     unsigned ndep; //!< Number of depsotions in hit
     float nelec; //!< Number of electrons in hit
-    float e; //!< energy in hit
-    float pitch; //!< Track pitch for hit
-    float rr; //!< Track residual range for hit
+    float e; //!< energy in hit [MeV]
+    float pitch; //!< Track pitch for hit [cm]
+    float rr; //!< Track residual range for hit [cm]
     int itraj; //!< Index of hit along trajectory
-    Vector3D p; //!< Location of hit
+    Vector3D p; //!< Location of hit [cm]
     float time; //!< Time of hit [ticks]
 
     TrueHit():
@@ -125,7 +125,10 @@ namespace sbn {
       ndep(0),
       nelec(0.),
       e(0.),
-      pitch(0.)
+      pitch(0.),
+      rr(0.),
+      itraj(-1),
+      time(0.)
       {
         // set the location to 0.
         p.x = 0;
--- a/sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
+++ b/sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
@@ -195,8 +195,6 @@ namespace sbn {
     Vector3D dir; //!< Direction of track 
     float length; //!< Length of track [cm]
 
-    unsigned ndaughters; //!< Number of daughters of track
-
     float hit_min_time_p0_tpcE; //!< Min hit time of track on plane 0 TPC E
     float hit_max_time_p0_tpcE; //!< Max hit time of track on plane 0 TPC E
     float hit_min_time_p1_tpcE; //!< Min hit time of track on plane 1 TPC E
@@ -222,6 +220,10 @@ namespace sbn {
     int selected; //!< Index of the tool that selected this track
     int nprescale; //!< Prescale of the tool that selected this track
 
+    std::vector<int> daughter_pdg; //!< Pandora PDG codes of daughter PFP's
+    std::vector<unsigned> daughter_nsp; //!< Number of space points in each daughter
+    std::vector<float> daughter_sp_toend_dist; //!< Smallest distance from any daughter Space Point to Track End [cm]
+
     std::vector<float> tracks_near_end_dist; //!< List of tracks near the end of this track
     std::vector<float> tracks_near_end_costh; //!< List of tracks near the end of this track
 
@@ -235,7 +237,6 @@ namespace sbn {
       cryostat(-1),
       clear_cosmic_muon(false),
       length(-1),
-      ndaughters(0),
       hit_min_time_p0_tpcE(-100000),
       hit_max_time_p0_tpcE(-100000),
       hit_min_time_p1_tpcE(-100000),
--- a/sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
+++ b/sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
@@ -105,8 +105,8 @@ namespace sbn {
     int16_t cryo; //!< Cryostat of hit
     int16_t tpc; //!< TPC of hit
     int16_t plane; //!< Plane of hit
-    int16_t wire; //!< Wire of hit
-    int16_t channel; //!< Channel of hit
+    int wire; //!< Wire of hit
+    int channel; //!< Channel of hit
 
     unsigned ndep; //!< Number of depositions in hit
     float nelec; //!< Number of electrons in hit
--- a/sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
+++ b/sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
@@ -98,6 +98,34 @@ namespace sbn {
      iproc(-1) {}
   };
 
+  struct TrueHit {
+    int16_t cryo; //!< Cryostat of hit
+    int16_t tpc; //!< TPC of hit
+    int16_t plane; //!< Plane of hit
+    int wire; //!< Wire of hit
+
+    unsigned ndep; //!< Number of depsotions in hit
+    float nelec; //!< Number of electrons in hit
+    float e; //!< energy in hit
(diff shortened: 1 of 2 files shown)


sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
```diff
@@ -98,6 +101,39 @@ namespace sbn {
...
+    int channel; //!< Channel of hit
+
+    unsigned ndep; //!< Number of depsotions in hit
+    float nelec; //!< Number of electrons in hit
+    float e; //!< energy in hit
+    float pitch; //!< Track pitch for hit
+    float rr; //!< Track residual range for hit
+    int itraj; //!< Index of hit along trajectory
+    Vector3D p; //!< Location of hit
+    float time; //!< Time of hit [ticks]
```
PetrilloAtWork, line 118: I _recommend_ that this class be made a "plain old data" structure by removing the explicit default constructor and initializing the values inline ([`[CF.155]`](https://sbnsoftware.github.io/sbn/codingconv/CodingConventionsExplained.html#CF155)):
```suggestion
  struct TrueHit {
    int16_t cryo { int16_t(-1) }; //!< Cryostat of hit
    int16_t tpc { int16_t(-1) }; //!< TPC of hit
    int16_t plane { int16_t(-1) }; //!< Plane of hit
    int wire { -1 }; //!< Wire of hit
    int channel { -1 }; //!< Channel of hit

    unsigned ndep { 0 }; //!< Number of depsotions in hit
    float nelec { 0.0 }; //!< Number of electrons in hit
    float e { 0.0 }; //!< energy in hit
    float pitch { 0.0 }; //!< Track pitch for hit
    float rr { 0.0 }; //!< Track residual range for hit
    int itraj { -1 }; //!< Index of hit along trajectory
    Vector3D p; //!< Location of hit
    float time { 0.0 }; //!< Time of hit [ticks]
```
Besides the advantages at language level, in this way you don't miss any initialization when you add an element.
I also _recommend_ the same treatment for `Vector3D`:
```cpp
  struct Vector3D {
    float x { std::numeric_limits<float>::signaling_NaN() };
    float y { std::numeric_limits<float>::signaling_NaN() };
    float z { std::numeric_limits<float>::signaling_NaN() };
  };
```
which will allow the above initialization to be written as `Vector3D p { 0.0, 0.0, 0.0 };` (that is one noticeable advantage of having no constructor specified). Otherwise, you will still need a constructor like
```cpp
    TrueHit()
      {
        // set the location to 0.
        p.x = 0;
        p.y = 0;
        p.z = 0;
      }
```

sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
```diff
@@ -136,6 +172,10 @@ namespace sbn {
     int   start_process; //!< Start G4 process of the particle. Values defned as enum in StandardRecord
     int   end_process; //!< End G4 process of the particle. Values defined as enum in StandardRecord
 
+    std::vector<TrueHit> truehits0; //!< List of True "hits" of this particle on Plane 0
+    std::vector<TrueHit> truehits1; //!< List of True "hits" of this particle on Plane 1
+    std::vector<TrueHit> truehits2; //!< List of True "hits" of this particle on Plane 2
```
PetrilloAtWork, line 177: I assume (and believe) that this is a deliberate and informed choice and that `0` really means "plane 0" and not e.g. view U or "best plane" etc.
But I would like a confirmation, just to be sure.

sbnobj/Common/Calibration/classes_def.xml
```diff
@@ -5,9 +5,15 @@
   <class name="sbn::TrackTruth" ClassVersion="10">
    <version ClassVersion="10" checksum="2749399767"/>
   </class>
-  <class name="sbn::TrueParticle" ClassVersion="10">
+  <class name="sbn::TrueParticle" ClassVersion="12">
+   <version ClassVersion="12" checksum="855399631"/>
+   <version ClassVersion="11" checksum="3478510212"/>
```
PetrilloAtWork, line 10: Please remove transient versions (if you implement some of the changes I have recommended, there will be more).
```suggestion
  <class name="sbn::TrueParticle" ClassVersion="11">
   <version ClassVersion="11" checksum="855399631"/>
```

sbnobj/Common/Calibration/classes_def.xml
```diff
@@ -5,9 +5,15 @@
...
    <version ClassVersion="10" checksum="2150513055"/>
   </class>
+  <class name="sbn::TrueHit" ClassVersion="11">
+   <version ClassVersion="11" checksum="118230262"/>
+   <version ClassVersion="10" checksum="1579485728"/>
+  </class>
```
PetrilloAtWork, line 16: As above: remove development versions.
```suggestion
  <class name="sbn::TrueHit" ClassVersion="10">
   <version ClassVersion="10" checksum="118230262"/>
  </class>
```

sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
```diff
@@ -98,6 +101,39 @@ namespace sbn {
...
+      ndep(0),
+      nelec(0.),
+      e(0.),
+      pitch(0.)
+      {
+        // set the location to 0.
+        p.x = 0;
+        p.y = 0;
+        p.z = 0;
+      }
```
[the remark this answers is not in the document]
gputnam: Fixed.

sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
gputnam: That is correct.

sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
```diff
@@ -98,6 +101,39 @@ namespace sbn {
      iproc(-1) {}
   };
 
+  struct TrueHit {
+    int16_t cryo; //!< Cryostat of hit
+    int16_t tpc; //!< TPC of hit
+    int16_t plane; //!< Plane of hit
+    int wire; //!< Wire of hit
+    int channel; //!< Channel of hit
```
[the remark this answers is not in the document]
gputnam: Fixed -- but kept as signed integer to allow for -1 as a "bad" value

sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
```diff
@@ -98,6 +101,39 @@ namespace sbn {
...
+
+    unsigned ndep; //!< Number of depsotions in hit
+    float nelec; //!< Number of electrons in hit
+    float e; //!< energy in hit
```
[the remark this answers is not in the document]
gputnam: Fixed

sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
gputnam: So -- I disagree with this convention. Maybe we could discuss at an Analysis Framework meeting.

This is using a C++ feature which is pretty unknown to most people, and most existing data product involve an explicit constructor. 

I can understand that there are some benefits, but I think it should be left up to the developer as to whether they are necessary.

gputnam:
Ok! @PetrilloAtWork @miquelnebot I believe I have addressed all comments on this PR as well.

sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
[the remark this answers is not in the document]
PetrilloAtWork: This will not do for `channel`, because there may be more than `32767` channels in the detector.

sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
PetrilloAtWork: This issue is important enough for maintainability for me to insist on the point.
Let's discuss at a meeting then.

sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
```diff
@@ -98,6 +101,42 @@ namespace sbn {
      iproc(-1) {}
   };
 
+  struct TrueHit {
+    int16_t cryo; //!< Cryostat of hit
+    int16_t tpc; //!< TPC of hit
+    int16_t plane; //!< Plane of hit
+    int16_t wire; //!< Wire of hit
+    int16_t channel; //!< Channel of hit
```
PetrilloAtWork, line 109: `channel` needs to accommodate more than `32767` channels:
```suggestion
    uint16_t channel; //!< Channel of hit
```
or set it to `uint32_t` (or simply `unsigned int`, if `raw::ChannelID` is not desired).
(note that such change may trigger another version in `classes_defs.xml`, which would then need to be trimmed again of the intermediate version)

sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
```diff
@@ -98,6 +101,42 @@ namespace sbn {
...
+      itraj(-1),
+      time(0.)
+      {
+        // set the location to 0.
+        p.x = 0;
+        p.y = 0;
+        p.z = 0;
+      }
+  };
+
```
PetrilloAtWork, line 139: As from in the previous review iteration, I think this to be an issue important enough for maintainability to insist to the C++11-style initialization pattern.
Let's discuss this in a meeting, then.

PetrilloAtWork (changes requested):
There is an issue introduced with the last review iteration that needs to be addressed.
Plus, there is an issue to be discussed live to reach an agreement.

sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
gputnam: I would like to be able to set the channel to -1 to allow for a "bad" value. I can revert the type to an int32 in that case.

sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
gputnam: @PetrilloAtWork -- if this needs a longer conversation. Can we merge in this PR and fix it if necessary down the line?

sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
PetrilloAtWork: Yes, or simply `int`.

wesketchum:
@PetrilloAtWork and @gputnam are we resolved on the review and can close this out?

PetrilloAtWork:
As much as I can see, the only stopping issue is the size of `channel` variable (signed 16 bits, unable to accommodate the full range of ICARUS TPC channel IDs). One other issue is pending but it should be discussed more extensively and we agreed we wouldn't stop this pull request because of it.

gputnam:
Hi all. I have changed the channel type back to accommodate the necessary range (https://github.com/SBNSoftware/sbnobj/pull/29/commits/54eafc6c3bed2f79fbdff55356ebcd096677c5d9). So this should be ready.

PetrilloAtWork (approved):

PR #32, opened 2021-11-11
Created PMT Hardware Trigger simulation producer

eyandel:
Definitions for the data product created by the pmtTriggerProducer module in SBNSoftware/sbndcode#214

### DIFF
--- a/sbnobj/SBND/CMakeLists.txt
+++ b/sbnobj/SBND/CMakeLists.txt
@@ -1,2 +1,3 @@
 add_subdirectory(CRT)
-add_subdirectory(Commissioning) 
+add_subdirectory(Commissioning)
+add_subdirectory(Trigger) 
--- a/sbnobj/SBND/Trigger/CMakeLists.txt
+++ b/sbnobj/SBND/Trigger/CMakeLists.txt
@@ -1,0 +1,10 @@
+cet_make( 
+  LIBRARIES
+    cetlib_except
+  NO_DICTIONARY
+  )
+
+art_dictionary(DICTIONARY_LIBRARIES sbnobj_SBND_Trigger)
+
+install_headers()
+install_source()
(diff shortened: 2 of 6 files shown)


pgreen135:
@eyandel what is the purpose of `classes.h`? and same for the `pmtTrigger.cc` file, as this is empty

eyandel:
> @eyandel what is the purpose of `classes.h`? and same for the `pmtTrigger.cc` file, as this is empty

These files are required for everything to compile properly, but this data product is currently very basic and doesn't yet need functions to go in here. So they should be mostly empty (with just some include statements) and do need to be here anyway.

eyandel:
> @eyandel as per [SBNSoftware/sbndcode#214](https://github.com/SBNSoftware/sbndcode/pull/214) could you update this with some information? Are the two PRs required to go together? Is this required for the next production?

This is just the necessary definitions for the data product created by the pmtTriggerProducer module. I don't think it needs too much information and instructions since these scripts won't do anything without the module (which now has the more detailed information) and no one should really be changing these sbnobj files, they just need to exist to use the module or read it's outputs. So yes, the PR's should go together and both are needed for the module to work.

pgreen135:
> > @eyandel what is the purpose of `classes.h`? and same for the `pmtTrigger.cc` file, as this is empty
> 
> These files are required for everything to compile properly, but this data product is currently very basic and doesn't yet need functions to go in here. So they should be mostly empty (with just some include statements) and do need to be here anyway.

Is the intent that these would be updated and added to in the future as this work develops further? If so I am happy with this.

eyandel:
> > > @eyandel what is the purpose of `classes.h`? and same for the `pmtTrigger.cc` file, as this is empty
> > 
> > 
> > These files are required for everything to compile properly, but this data product is currently very basic and doesn't yet need functions to go in here. So they should be mostly empty (with just some include statements) and do need to be here anyway.
> 
> Is the intent that these would be updated and added to in the future as this work develops further? If so I am happy with this.

Yes definitely. This is just the first iteration.

pgreen135 (approved):

PR #33, opened 2021-11-15
Calibration NTuple Update

gputnam:
Refine truth matching variables; add in SCE corrected versions where useful; add in an addition Hit truth backtracking method.

### DIFF
--- a/sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
+++ b/sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
@@ -35,6 +35,11 @@ namespace sbn {
 
   };
 
+  struct HitTruth {
+    float e;
+    float nelec;
+  };
+
   struct HitInfo {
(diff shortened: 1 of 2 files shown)


wesketchum (changes requested):
Looks good, but one minor request: I think the previous version was '10' for `sbn::TrueHit` and `sbn::HitInfo`. checksums are different for ClassVersion10 for `sbn::TrueHit`: can we be sure to retain the original checksum, and then remove any intermediate ClassVersions that aren't needed (e.g. 11-13, and rename 14-->11 unless we've had releases/productions/etc. with some intermediate version). Similar for other ones. (think they don't have discrepant checksums, but some intermediate versions.)

gputnam:
Oh! Good catch.

I believe the `HitInfo` checksums are ok, but I did cleanup the `TrueHit` history.

wesketchum (approved):
Thanks! Looks good to me now.

PR #37, opened 2021-12-07
Feature/lg nutools v3 12 00

miquelnebot:


henrylay97 (approved):
Can't see why not, its been made by the larsoft team for this purpose! Might be worth a quick `trigger build` spin?

PR #39, opened 2021-12-15
adding association of anab::T0 to sbn::crt::CRTHit

biswaranjan91:


sbnobj/Common/CRT/classes_def.xml
```diff
@@ -1,6 +1,7 @@
 <lcgdict>
 
-  <class name="sbn::crt::CRTHit" ClassVersion="14">
+  <class name="sbn::crt::CRTHit" ClassVersion="15">
+   <version ClassVersion="15" checksum="3264936168"/>
```
PetrilloAtWork, line 4: While the pull request states that the dictionary for the association is added, this change suggests that the `sbn::crt::CRTHit` object is also changed. But this pull request is not adding any change to it, so its checksum (and version) should not change.
Please clarify.

sbnobj/Common/CRT/classes_def.xml
```diff
@@ -36,6 +37,10 @@
  <class name="art::Assns<sbn::crt::CRTHit, sbn::crt::CRTTrack,    void>"           />
  <class name="art::Wrapper< art::Assns<sbn::crt::CRTHit, sbn::crt::CRTTrack, void> >"           />
 
+ <class name="art::Assns<anab::T0,sbn::crt::CRTHit,void>" />
+ <class name="art::Assns<sbn::crt::CRTHit,anab::T0,void>" />
+ <class name="art::Wrapper<art::Assns<anab::T0,sbn::crt::CRTHit,void> >" />
+ <class name="art::Wrapper<art::Assns<sbn::crt::CRTHit,anab::T0,void> >" />
```
PetrilloAtWork, line 43: Tradition demanded also the _art_ pointers to the objects (here, `art::Ptr<sbn::crt::CRTHit>`, since `anab::T0`'s one is defined in `lardataobj`) and their `std::pair` (involved in `art::Assns`) and their `std::vector`s to be added.
I believe I got away without that in the past, so I just recommend that a full test of write and read back is performed to validate this choice.

sbnobj/ICARUS/CRT/classes_def.xml
```diff
@@ -8,7 +8,7 @@
 
   <class name="std::map< uint8_t, uint16_t >"/>
   <class name="std::map< unsigned char, std::vector< std::pair<int,float> > > "/>
-
+  <class name="std::pair<unsigned char,std::vector<std::pair<int,float> > > "/>
```
PetrilloAtWork, line 11: Where does the need for this come from?

PetrilloAtWork (changes requested):
A clarification is needed on the change of checksum of `crt::sbn::CRTHit`.
Also the confirmation that a full write/read cycle was successfully completed would be good.

sbnobj/Common/CRT/CRTHit.hh
```diff
@@ -26,12 +26,12 @@ namespace sbn::crt {
...
-      uint32_t ts0_s;
+      uint64_t ts0_s;
       int8_t ts0_s_corr;
       
-      uint32_t ts0_ns;
-      int32_t ts0_ns_corr;
-      int32_t ts1_ns;
+      uint64_t ts0_ns;
+      int64_t ts0_ns_corr;
+      int64_t ts1_ns;
```
PetrilloAtWork, line 34: Could you explain the need for this change?
There is no inline documentation, so I can only guess from the data member names: the `_ns` ones represent the number of nanoseconds of a timestamp since the last round second. So these are in the `[ 0 ; 1'000'000'000 [` range.
I _suppose_ that the `_s` part is the number of seconds since some Epoch, so unless you pick as Epoch a day in the 15th century, 32 bit should suffice (they span 68 years if signed, twice that much if unsigned). Ans in no case the nanosecond part requires 64 bit, unless it is _the full timestamp_ expressed in nanoseconds.
If 64-bit timestamps are desired, I'd rather recommend two member functions:
```cpp
      int64_t ts0() const { return static_const<int64_t>(ts0_s) * 1'000'000'000LL + ts0_ns; }
      int64_t ts1() const { return static_const<int64_t>(ts0_s) * 1'000'000'000LL + ts1_ns; }
```
(I did not apply the "correction" `ts0_s_corr` because I don't know if one should here).

sbnobj/Common/CRT/classes_def.xml
biswaranjan91: https://github.com/SBNSoftware/sbnobj/commit/d7c9ab34265bc902b795c951a716e346e2d78738#diff-fdbc5cb8037ead542ffc2d38998a6785bc87d5eb1eb58c8a5e28ee5a2f5ab2a5

I have added this in to the CRTHIt.hh

PetrilloAtWork (commented):
I exercise much stricter care than average when reviewing data products, because once they are in the data files it may be really hard to "fix" them. Hence my comments:

1. The changes on `CRTHit` explain now the change in checksum, but I would like some clarifications about those changes, as noted in the relevant comment.
2. I strongly question the presence of a `std::map` in a data product: they waste a lot of memory and they make the creation of the data product slow. I would gladly discuss alternative approaches with you or with the proponent/maintainer of this class.
3. I would _really_ appreciate inline documentation reporting the meaning and format of the object and its data; in particular, `pesmap` is completely obscure to me.

These last two points go beyond the review of this pull request and while I hope in some follow up they don't prevent the request from moving forward. The first one should instead be addressed.

sbnobj/ICARUS/CRT/classes_def.xml
biswaranjan91: no dictionary for class pair<unsigned char,vector<pair<int,float> > > is available

I was seeing above messages during running the CRTHit reconstruction, adding this line helps. Honestly did not made any effort where it was originating. I will give you more details on it.

PetrilloAtWork: With the new commit I now see the definition of `CRTHit`, and I think this `std::pair` may be necessary as an internal component of `std::map< unsigned char, std::vector< std::pair<int,float> > >`. Now I am just surprised that it was not caught earlier, since it's pre-existing this pull request, but this is now clarified..

sbnobj/Common/CRT/CRTHit.hh
biswaranjan91: Answer to 1 : T0 is an absolute timestamp, i.e. number of ns since beginning of 1970. A big number. It is also corrected for the fact that PPS takes time to arrive to various FEBs. In CRTData we have assigned T0 to uint64_t, however we pass this to CRTHit it was assigned to uint32_t, So there is a need of change in the CRTHit. I am not sure the use of `ts0_s_corr` I am trying to figure it out, and revisiting the CRTHitRecoalg, will have more information in day or two and love to discuss if there are better ways to improve. Thanks a lot for helping/guiding me.

sbnobj/ICARUS/CRT/classes_def.xml
biswaranjan91: Thank you.

sbnobj/Common/CRT/classes_def.xml
biswaranjan91: It seems working and I am confirming it in this way. Maybe I will try adding more in next release.

sbnobj/Common/CRT/CRTHit.hh
PetrilloAtWork: If `ts0_ns` are the full timestamps, then there should be no need for `ts0_s`, which is just `ts0_ns / 1'000'000'000`.
And anyway `ts0_s` should not need 64 bit if it's only the number of seconds.
Can you show me a producer module where this data is filled?

sbnobj/Common/CRT/classes_def.xml
PetrilloAtWork: Thank you.

sbnobj/Common/CRT/CRTHit.hh
biswaranjan91: Here is the Algorithm accessing these variables.
https://github.com/SBNSoftware/icaruscode/blob/develop/icaruscode/CRT/CRTUtils/CRTHitRecoAlg.cc#L255

Yes, you are correct, as we have ts0_ns, it can be extracted.  As it was there for long time back and used by both detector, we have not changed this variable. It will be better to be removed.

PetrilloAtWork: Thank you for the pointer to the code, it is much clearer now.
The primary source of my confusion is that in other data products when there are two fields like `t0_s` and `t0_ns`, the first contains only the seconds of the time stamp from the Epoch, and the second only the nanoseconds after that second, so the full timestamp is `t0_s * 1'000'000'000ULL + t0_ns` (and note that I need to specify that the 10⁹ factor is a 64-bit number, unsigned long long, with `ULL` suffix).
If you have reasons to nevertheless keep both the absolute timestamp (`t0_ns`) _and_ the second-only part (`t0_s`), so ahead with it. But please add inline documentation explaining all of these conventions.
For example:
```cpp
      uint64_t ts0_s;  ///< Second-only part of timestamp T0.
      int8_t ts0_s_corr; ///< [what is this??]

      uint64_t ts0_ns;  ///< Timestamp T0 (from White Rabbit), in UTC absolute time scale in nanoseconds from the Epoch.
      int64_t ts0_ns_corr; ///< [what is this??]
      int64_t ts1_ns; ///< Timestamp T1 ([?]), in UTC absolute time scale in nanoseconds from the Epoch.
```
Also note that the code you pointed out has a minor issue: the conversions in seconds unnecessarily convert the number in `double` and then back to `int` because `1e-9` is a double (should go with `/ 1'000'000'000` instead); this is probably not fatal because there should still be no loss of precision _in the second part_ of the timestamp.

sbnobj/Common/CRT/CRTHit.hh
```diff
@@ -22,30 +22,33 @@ namespace sbn::crt {
...
-      std::map< uint8_t, std::vector<std::pair<int,float> > > pesmap;
-      float peshit;
+      std::vector<uint8_t> feb_id; ///< FEB address 
+      std::map< uint8_t, std::vector<std::pair<int,float> > > pesmap; ///< Saves signal hit information (FEB, local-channel and PE) .
+      float         peshit; ///< Total photo-electron (PE) in a crt hit.
      
-      uint32_t ts0_s;
-      int8_t ts0_s_corr;
+      uint64_t       ts0_s; ///< Second-only part of timestamp T0.
+      int8_t    ts0_s_corr; ///< [Honestly, not sure at this point, it was there since long time (BB)]
```
PetrilloAtWork, line 30: > [Honestly, not sure at this point, it was there since long time (BB)]

:rofl:

PetrilloAtWork (approved):
Thank you for adding the documentation of the data.
I always question the use of something as complex as `std::map` in data products, but this pull request did not introduce it so it's not on this table.
Good to go for me.

PR #45, opened 2022-03-10
NuMI Beamline Accounting

gputnam:
Add in more information to NuMI object for beamline accounting. Intended for production release.

### DIFF
--- a/sbnobj/Common/POTAccounting/classes_def.xml
+++ b/sbnobj/Common/POTAccounting/classes_def.xml
@@ -6,7 +6,7 @@
    <version ClassVersion="10" checksum="13541"/>
   </class>
   <class name="sbn::NuMISpillInfo" ClassVersion="13" >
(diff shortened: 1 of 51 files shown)


jzennamo (approved):

sbnobj/Common/POTAccounting/classes_def.xml
```diff
@@ -5,13 +5,22 @@
    <version ClassVersion="11" checksum="3354055579"/>
    <version ClassVersion="10" checksum="13541"/>
   </class>
+  <class name="sbn::NuMISpillInfo" ClassVersion="13" >
```
wesketchum, line 8: This is a new object, correct? If so, I think that it'd be better to change ClassVersion13 to 10, and remove the older ones. No need to keep versioning around for what you used as you developed.

wesketchum:
Marking as approved, but @gputnam , please see comment above on the checksums.

PR #47, opened 2022-03-16
CRUMBS

henrylay97:
Create the CRUMBSResult object to store the score and input variables from CRUMBS's BDT.

Links to main PR https://github.com/SBNSoftware/sbncode/pull/250

### DIFF
--- a/sbnobj/Common/Reco/CRUMBSResult.cc
+++ b/sbnobj/Common/Reco/CRUMBSResult.cc
@@ -1,1 +1,56 @@
 #include "sbnobj/Common/Reco/CRUMBSResult.h"
+
+sbn::CRUMBSResult::CRUMBSResult(float score, float tpc_CRFracHitsInLongestTrack, float tpc_CRLongestTrackDeflection, float tpc_CRLongestTrackDirY,
+				float tpc_CRNHitsMax, float tpc_NuEigenRatioInSphere, float tpc_NuNFinalStatePfos, float tpc_NuNHitsTotal,
+				float tpc_NuNSpacePointsInSphere, float tpc_NuVertexY, float tpc_NuWeightedDirZ, float tpc_StoppingChi2CosmicRatio,
+				float pds_FMTotalScore, float pds_FMPE, float pds_FMTime, float crt_TrackScore, float crt_HitScore,
+				float crt_TrackTime, float crt_HitTime)
+: score(score)
+  , tpc_CRFracHitsInLongestTrack(tpc_CRFracHitsInLongestTrack)
+  , tpc_CRLongestTrackDeflection(tpc_CRLongestTrackDeflection)
+  , tpc_CRLongestTrackDirY(tpc_CRLongestTrackDirY)
+  , tpc_CRNHitsMax(tpc_CRNHitsMax)
+  , tpc_NuEigenRatioInSphere(tpc_NuEigenRatioInSphere)
+  , tpc_NuNFinalStatePfos(tpc_NuNFinalStatePfos)
+  , tpc_NuNHitsTotal(tpc_NuNHitsTotal)
+  , tpc_NuNSpacePointsInSphere(tpc_NuNSpacePointsInSphere)
+  , tpc_NuVertexY(tpc_NuVertexY)
+  , tpc_NuWeightedDirZ(tpc_NuWeightedDirZ)
+  , tpc_StoppingChi2CosmicRatio(tpc_StoppingChi2CosmicRatio)
+  , pds_FMTotalScore(pds_FMTotalScore)
+  , pds_FMPE(pds_FMPE)
+  , pds_FMTime(pds_FMTime)
+  , crt_TrackScore(crt_TrackScore)
+  , crt_HitScore(crt_HitScore)
+  , crt_TrackTime(crt_TrackTime)
+  , crt_HitTime(crt_HitTime)
+{
+}
+
+sbn::CRUMBSResult::CRUMBSResult(float score)
+  : score(score)
+{
+}
+
+sbn::CRUMBSResult::CRUMBSResult()
+  : score(-5.f)
+  , tpc_CRFracHitsInLongestTrack(-5.f)
+  , tpc_CRLongestTrackDeflection(-5.f)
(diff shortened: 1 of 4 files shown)


sbnobj/Common/Reco/CRUMBSResult.h
```diff
@@ -0,0 +1,105 @@
...
+      , pds_FMPE(pds_FMPE)
+      , pds_FMTime(pds_FMTime)
+      , crt_TrackScore(crt_TrackScore)
+      , crt_HitScore(crt_HitScore)
+      , crt_TrackTime(crt_TrackTime)
+      , crt_HitTime(crt_HitTime)
+    {
+    }
+
+    float score;
```
etyley, line 83: Documenting the units would be useful (where applicable)

sbnobj/Common/Reco/CRUMBSResult.h
```diff
@@ -0,0 +1,105 @@
...
+    float tpc_CRFracHitsInLongestTrack;
+    float tpc_CRLongestTrackDeflection;
+    float tpc_CRLongestTrackDirY;
+    float tpc_CRNHitsMax;
+    float tpc_NuEigenRatioInSphere;
+    float tpc_NuNFinalStatePfos;
+    float tpc_NuNHitsTotal;
```
etyley, line 90: It seems like some of these should be `nts?

etyley (approved):
Minor comments about documentation, but no showstoppers

sbnobj/Common/Reco/CRUMBSResult.h
```diff
@@ -0,0 +1,105 @@
...
+      , pds_FMPE(-5.f)
+      , pds_FMTime(-5.f)
+      , crt_TrackScore(-5.f)
+      , crt_HitScore(-5.f)
+      , crt_TrackTime(-5.f)
+      , crt_HitTime(-5.f)
+    {
+    }
+
+    CRUMBSResult(const float score, const float tpc_CRFracHitsInLongestTrack, const float tpc_CRLongestTrackDeflection, const float tpc_CRLongestTrackDirY,
```
cjbackhouse, line 56: Personally I don't think all these `consts` help in such a situation.

sbnobj/Common/Reco/CRUMBSResult.h
```diff
@@ -0,0 +1,105 @@
...
+      , tpc_NuVertexY(-5.f)
+      , tpc_NuWeightedDirZ(-5.f)
+      , tpc_StoppingChi2CosmicRatio(-5.f)
+      , pds_FMTotalScore(-5.f)
+      , pds_FMPE(-5.f)
+      , pds_FMTime(-5.f)
+      , crt_TrackScore(-5.f)
+      , crt_HitScore(-5.f)
+      , crt_TrackTime(-5.f)
+      , crt_HitTime(-5.f)
```
cjbackhouse, line 52: Can this be
```suggestion
      : CRUMBResult(),
        score(score)
```

cjbackhouse (approved):

sbnobj/Common/Reco/CRUMBSResult.cc
```diff
@@ -0,0 +1 @@
+#include "sbnobj/Common/Reco/CRUMBSResult.h"
```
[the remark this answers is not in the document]
henrylay97: This directory seems to contain almost all `.cc` so I was just sticking with convention, happy to move the constructors across :)

sbnobj/Common/Reco/CRUMBSResult.h
henrylay97: Okay, will remove

sbnobj/Common/Reco/CRUMBSResult.h
henrylay97: So I had a bit of back and forth on this. Pandora creates all the mva features as doubles (even for the clear ints like NHits), TMVA only accepts floats not doubles so I opted for casting them all to floats and not worrying about casting back to an int and the possible pitfalls there. Happy to amend that if people don't agree.

cjbackhouse: I think the ones that are logically int should be so on our side, and we'll just convert when pulling things from pandora, and writing things to TMVA. Just because all the world around is mad doesn't mean you have to let it infect the part you control. One day maybe pandora will be able to store more types. `std::round()` should be pretty safe.

henrylay97: I have gone with your suggestion. Have kept them as floats in the CRUMBS_module as they are associated with the TMVA tree which requires floats. I then use std::round() to cast them to ints as we make the CRUMBSResult object and they remain as ints later in the CAF element too.

sbnobj/Common/Reco/CRUMBSResult.h
henrylay97: Kind of, have to change the order and then it calls it implicitly. I definitely learned a few things looking at this :D

sbnobj/Common/Reco/classes_def.xml
```diff
@@ -51,6 +51,18 @@
   <class name="art::Assns<sbn::SimpleFlashMatch,recob::PFParticle, void>" />
   <class name="art::Wrapper<art::Assns<sbn::SimpleFlashMatch, recob::PFParticle, void> >" />
 
+  <class name="sbn::CRUMBSResult" ClassVersion="15">
+   <version ClassVersion="15" checksum="1001500335"/>
+   <version ClassVersion="14" checksum="3636907335"/>
+  </class>
```
cjbackhouse, line 57: Could collapse these down to one version again, and label it version 10.

sbnobj/Common/Reco/CRUMBSResult.h
cjbackhouse: Are you sure? I'm pretty sure the score-only constructor you have right now will leave the other fields initialized with junk. Did my suggestion not compile?

Actually, I'm now remembering what Gianluca would say about all this, which would be to recommend you give these default values inline in the header file. Then they do implicitly apply to all constructors.

PR #52, opened 2022-04-28
Change Time Data Type for CRTHit and CRTTrack

### LINKED ISSUE #51: Time Data Type for CRTHit and CRTTrack
Hello! I noticed that the data products for CRT hits and tracks contain times in unsigned int format. But hit times come from averaging the times of two X and Y modules, and also the time of each strip comes from averaging the two SiPMs. I would propose we change the type to `float`.

Also, the fact that they are unsigned I think it's a problem. We may want to correct for time offsets at the reconstruction stage, which will make some hit times negative.

Please, let me know if changing to `float` is an issue, otherwise, I can work on this modification. Tagging here all relevant people, my apologies if I am missing anyone, please feel free to add. @ilepetic @Umut-Kose @mstancar @biswaranjan91 

**CRTHit**

https://github.com/SBNSoftware/sbnobj/blob/aed8815534c160e68943271ce6f147416e89d5fe/sbnobj/Common/CRT/CRTHit.hh#L29-L34

**CRTTrack**

https://github.com/SBNSoftware/sbnobj/blob/aed8815534c160e68943271ce6f147416e89d5fe/sbnobj/Common/CRT/CRTTrack.hh#L26-L31

marcodeltutto:
As discussed in issue #51, this PR changes the data type for time variables from `unsigned int` to `double` for both CRTHits and CRTTracks. Resolves #51.

The main change is in commit [393f861](https://github.com/SBNSoftware/sbnobj/pull/52/commits/393f8610371f0903303c8874779dd75917bbab68).

### DIFF
--- a/sbnobj/Common/CRT/CRTHit.hh
+++ b/sbnobj/Common/CRT/CRTHit.hh
@@ -27,11 +27,11 @@ namespace sbn::crt {
       float         peshit; ///< Total photo-electron (PE) in a crt hit.
      
       uint64_t       ts0_s; ///< Second-only part of timestamp T0.
-      int8_t    ts0_s_corr; ///< [Honestly, not sure at this point, it was there since long time (BB)]
+      double    ts0_s_corr; ///< [Honestly, not sure at this point, it was there since long time (BB)]
       
-      uint64_t      ts0_ns; ///< Timestamp T0 (from White Rabbit), in UTC absolute time scale in nanoseconds from the Epoch.
-      int8_t   ts0_ns_corr; ///< [Honestly, not sure at this point, it was there since long time (BB)]
-      int64_t       ts1_ns; ///< Timestamp T1 ([signal time w.r.t. Trigger time]), in UTC absolute time scale in nanoseconds from the Epoch.
+      double        ts0_ns; ///< Timestamp T0 (from White Rabbit), in UTC absolute time scale in nanoseconds from the Epoch.
+      double   ts0_ns_corr; ///< [Honestly, not sure at this point, it was there since long time (BB)]
+      double        ts1_ns; ///< Timestamp T1 ([signal time w.r.t. Trigger time]), in UTC absolute time scale in nanoseconds from the Epoch.
      
       int            plane; ///< Name of the CRT wall (in the form of numbers).
 
@@ -46,8 +46,8 @@ namespace sbn::crt {
 
       CRTHit() {}
 
-      int64_t ts0() const { return static_cast<int64_t>(ts0_s) * 1'000'000'000LL + ts0_ns; }
-      int64_t ts1() const { return static_cast<int64_t>(ts0_s) * 1'000'000'000LL + ts1_ns; }
+      int64_t ts0() const { return static_cast<int64_t>(ts0_s) * 1'000'000'000LL + static_cast<int64_t>(ts0_ns); }
+      int64_t ts1() const { return static_cast<int64_t>(ts0_s) * 1'000'000'000LL + static_cast<int64_t>(ts1_ns); }
 
     };
 
--- a/sbnobj/Common/CRT/CRTHit.hh
+++ b/sbnobj/Common/CRT/CRTHit.hh
@@ -22,17 +22,17 @@ namespace sbn::crt {
 
     struct CRTHit{
 
-      std::vector<uint8_t> feb_id; ///< FEB address 
+      std::vector<uint8_t> feb_id; ///< FEB address
       std::map< uint8_t, std::vector<std::pair<int,float> > > pesmap; ///< Saves signal hit information (FEB, local-channel and PE) .
       float         peshit; ///< Total photo-electron (PE) in a crt hit.
-     
+
       uint64_t       ts0_s; ///< Second-only part of timestamp T0.
       double    ts0_s_corr; ///< [Honestly, not sure at this point, it was there since long time (BB)]
-      
+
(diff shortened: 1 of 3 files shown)


mstancar (approved):

sbnobj/Common/CRT/CRTHit.hh
```diff
@@ -22,17 +22,17 @@ namespace sbn::crt {
...
-     
+
       uint64_t       ts0_s; ///< Second-only part of timestamp T0.
-      int8_t    ts0_s_corr; ///< [Honestly, not sure at this point, it was there since long time (BB)]
-      
-      uint64_t      ts0_ns; ///< Timestamp T0 (from White Rabbit), in UTC absolute time scale in nanoseconds from the Epoch.
-      int8_t   ts0_ns_corr; ///< [Honestly, not sure at this point, it was there since long time (BB)]
-      int64_t       ts1_ns; ///< Timestamp T1 ([signal time w.r.t. Trigger time]), in UTC absolute time scale in nanoseconds from the Epoch.
-     
+      double    ts0_s_corr; ///< [Honestly, not sure at this point, it was there since long time (BB)]
```
aaduszki, line 30: Since you are working on this, could someone actually clarify what is ts0_s_corr and ts1_ns_corr? Changing the data format is always difficult, so in case further changes are needed, it's more efficient to include them all at once.

sbnobj/Common/CRT/CRTTrack.hh
```diff
@@ -17,21 +17,21 @@
 #include <map>
 
 namespace sbn::crt {
-  
+
   struct CRTTrack{
 
     std::vector<uint8_t> feb_id;
     std::map< uint8_t, std::vector<std::pair<int,float> > > pesmap;
     float peshit;
```
aaduszki, line 25: I suggest to add comments with descriptions to all variables. Or even better, rename variables so that they don't use obscure abbreviation, and instead are can be understood more intuitively.

aaduszki (commented):
@biswaranjan91 commenting since you requested my review.
This update allows the timestamps to store sub-nanosecond values. I really wish the FEBs to produce adequately precise data, but let's hope for the best.
The update will require corresponding changes in the analysis code. I don't really work with that code, so perhaps someone who does should comment. Myself I don't see anything wrong in this implementation.
I wrote some minor suggestions, but they are not a show stopper.

biswaranjan91 (approved):

sbnobj/Common/CRT/CRTTrack.hh
marcodeltutto: I agree, I added comments with description, and removed the two CRTTrack variables `feb_id` and `pesmap`which are used neither in `sbndcode` nor in `icaruscode`.

sbnobj/Common/CRT/CRTHit.hh
marcodeltutto: I tried to ask around in the past days but could find clarification on `ts0_s_corr` and `ts1_ns_corr`.

mstancar:
From: Marco Del Tutto ***@***.***>
Reply-To: SBNSoftware/sbnobj ***@***.***>
Date: Friday, April 29, 2022 at 3:41 PM
To: SBNSoftware/sbnobj ***@***.***>
Cc: Michelle D Stancari ***@***.***>, Review requested ***@***.***>
Subject: Re: [SBNSoftware/sbnobj] Change Time Data Type for CRTHit and CRTTrack (PR #52)


@marcodeltutto commented on this pull request.

________________________________

In sbnobj/Common/CRT/CRTHit.hh<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_SBNSoftware_sbnobj_pull_52-23discussion-5Fr862137114&d=DwMCaQ&c=gRgGjJ3BkIsb5y6s49QqsA&r=XwT4z_xoYuahpWzd6VjM1ZO21Bag6v9xST-R_JW6GZM&m=jpSb8lPwqeHO36Q1SGs1Sp06E8Nx_Cnf2OYPuO3ufZSZpTVQfydfLWOMWDHtnMxt&s=MsmxJVQTww47-U_7lBtHfFmVdMGvcmTSpKyFmacA5Bg&e=>:

>        uint64_t       ts0_s; ///< Second-only part of timestamp T0.

-      int8_t    ts0_s_corr; ///< [Honestly, not sure at this point, it was there since long time (BB)]

-

-      uint64_t      ts0_ns; ///< Timestamp T0 (from White Rabbit), in UTC absolute time scale in nanoseconds from the Epoch.

This  (above) is supposed to only be the ns part of the timestamp – to get the time from the epoch you have to combine it with ts0_s.

-      int8_t   ts0_ns_corr; ///< [Honestly, not sure at this point, it was there since long time (BB)]

-      int64_t       ts1_ns; ///< Timestamp T1 ([signal time w.r.t. Trigger time]), in UTC absolute time scale in nanoseconds from the Epoch.

Ditto for this one – this is supposed to be only the ns part of the timestamp . . . .

-

+      double    ts0_s_corr; ///< [Honestly, not sure at this point, it was there since long time (BB)]

I tried to ask around in the past days but could find clarification on ts0_s_corr and ts1_ns_corr.

I think these two variables are intended to be  useful when constructing the UTC absolute time from the EPOCH from the ts0_ns and ts0_s values.  There are no timing offsets in simulation, so SBND is not currently using them.  But they could be useful when we have data

marcodeltutto:
trigger build

PR #56, opened 2022-05-27
Initialize fWidth

jedori0228:
For the multisigma modes where we fill `fWidths`, `fWidth` was not initialized and cause errors when we compare two objects like in https://github.com/SBNSoftware/sbncode/blob/d40d6334c05aaae67cb3ef3ebfb97ec18549773c/sbncode/CAFMaker/CAFMaker_module.cc#L469. This is updated, and also we have to check `fWidths` in  `operator==`.

### DIFF
--- a/sbnobj/Common/SBNEventWeight/EventWeightParameterSet.h
+++ b/sbnobj/Common/SBNEventWeight/EventWeightParameterSet.h
@@ -21,9 +21,7 @@ struct EventWeightParameter {
 
   /** Constructor specifying all parameter properties. */
   EventWeightParameter(std::string name, float mean, float width, size_t covIndex=0)
-      : fName(name), fMean(mean), fWidth(width), fCovIndex(covIndex) {
-    fWidths.clear();
-  }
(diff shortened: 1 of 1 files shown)


cjbackhouse (approved):

sbnobj/Common/SBNEventWeight/EventWeightParameterSet.h
```diff
@@ -21,10 +21,12 @@ struct EventWeightParameter {
 
   /** Constructor specifying all parameter properties. */
   EventWeightParameter(std::string name, float mean, float width, size_t covIndex=0)
-      : fName(name), fMean(mean), fWidth(width), fCovIndex(covIndex) {}
+      : fName(name), fMean(mean), fWidth(width), fCovIndex(covIndex) {
+    fWidths.clear();
```
cjbackhouse, line 25: This line is not necessary though. fWidths will automatically be empty when the object is constructed.

jedori0228: Ah, just removed this line to make it neater :)

PR #60, opened 2022-08-11
Add equivalent neutrino enrgy to MeVPrtl objects.

gputnam:


miquelnebot (approved):
Approve as SBNSoftware/sbncode#278

PR #72, opened 2022-11-30
Fix sbn::crt::ts1()

jedori0228:
This line was edited a while ago when T1 of CRTHit was not working properly. Now that we have correct numbers, and by the fact the T1 is a relative time not a timestamp, we only need ts1_ns to describe entire T1.


francescopoppi (commented):
Hi Jae! I think this change is needed and fine, the problem is that since this is common with SBND, we should also need an approval from SBND CRT experts.

francescopoppi (approved):

henrylay97 (approved):
Thanks for looping us in. I definitely approve this change, it fits our usage (and intended usage) of the t1 variable. 

Feel free to remove the SBND comment from it too :)

jedori0228:
Thank you @henrylay97 and @francescopoppi, I updated the comment and now it's ready to be go through ci test.

PR #76, opened 2023-03-03
Added sbn::ExtraTriggerInfo::triggerFromBeamGate()

PetrilloAtWork:
This is a simple addition to this data product interface.
The new member function returns the trigger time with respect to the beam gate time, and I added it because the conversions to signed are tedious if written correctly, so I'd rather do it once.

This is a gregarious request that should be merged only when some other thing is being merged in `sbnobj`: it's not worth its own release.
Also note that this is a GitHub edit — I trust on the C.I. test to tell if it does not compile.

I also let you choose whom to ask to review it (a.k.a. I have no idea).


jzettle (approved):
One line change looks good to me

PR #81, opened 2023-04-26
Update naming for ToF object

henrylay97:
Following extensive work to rewrite the CRT reconstruction for SBND the PR strategy has been agreed to be a series of PRs against a common branch (`feature/hlay_crt_clustering_base`) for the (inevitably very long) review stage. Please don't use CI tests yet. They will be used on the final PR when the common branch is compared to `develop`. At this stage the individual branches may not each compile on their own.

This PR just updates some naming in the ToF variable to ensure consistency with the new reco.


vmeddage44 (approved):
Agree with the changes.

PR #90, opened 2023-06-20
SBND OpT0Finder Objects

lynnt-uchicago:
created new OpT0Finder data products, these replace the T0 objects storing OpT0Finder results in `sbndcode`.


ggamezdiego (approved):

henrylay97 (approved):
Looks good, a couple of small queries but nothing that should hold this up.

PR #96, opened 2023-09-14
Additions to sbn::ExtraTriggerInfo data product to store ICARUS adder information

PetrilloAtWork:
I am adding some additional bits for the adders.
Adders are hardware performing analogue sum of PMT, and discriminating them, and the result can be integrated into the input used for the trigger hardware to take its decision.
This change is backward compatible via schema evolution; I have regularly used production ICARUS Stage0 data containing it, without any issue showing up.

Additional changes facilitate the access to these and other bits via Python interface.

Not sure whom to add for SBND review... @jzettle is the in-house expert of ICARUS trigger decoding, which fills this data product (and which will be updated in another dependent `icaruscode` pull request).
There is also a small change to `CMakeLists.txt` adopting a suggestion from the CMake output, which updates a now deprecated syntax.


lyates17:
I don't have the expertise to comment on whether this code is correct for the purpose of meeting ICARUS's needs (although it looks reasonable). I see no significant conflict with SBND's needs and thus have no objection to this being merged. An ICARUS expert should review and approve.

jzettle (approved):
I don't see anything clearly incorrect here. This is parallel with https://github.com/SBNSoftware/icaruscode/pull/631 and https://github.com/SBNSoftware/icarusalg/pull/73 so should be coordinated with those.

miquelnebot (approved):
https://github.com/SBNSoftware/sbnobj/pull/96#issuecomment-1750833459

PR #97, opened 2023-09-18
cnnscore

wjdanswjddl:
Create PFP-level objects for the CNN hit classification score


yangtj207 (approved):
The changes look good.

PR #100, opened 2023-11-01
Taking into account MinBias Triggers

jzennamo:
This is connected to:
[feature/zennamo_AccountingForMinBiasTriggers](https://github.com/SBNSoftware/sbncode/pull/394) (SBNSoftware/sbncode#394)


PetrilloAtWork (approved):
The data member needs to be real to accommodate statistical corrections that, event by event, may result into fractional gate counts. The actual gate count is still an integer.

PR #105, opened 2024-02-18
Add true trajectory points to truth information

brucehoward-physics:
Cherry-picking in what I think are the relevant commits from Gray's feature branch used in the reprocessing.

Not sure if this is necessary if we don't include the calibration ntuple change in icaruscode but it should be there as of now.


mrmooney (approved):
Looks fine to me

brucehoward-physics:
@miquelnebot I think the CI isn't setting up the right products. For this and https://github.com/SBNSoftware/sbncode/pull/437 I believe I am seeing some potential issues to this regard.

E.g. this has
```
sbnanaobj v09_20_06
```
when I think this version should be setting up
```
sbnanaobj v09_20_06_03
```

If we can build with this second release that might be better 🤞🤞

PR #106, opened 2024-02-29
Adding trigger information for ICARUS

PetrilloAtWork:
ICARUS is introducing a complementary trigger based on analogue sum of PMT signals.
This request extends `sbn::ExtraTriggerInfo` to accommodate for this new trigger:
* additional bits (`sectorStatus`) describe the primitive elements of a trigger based on the detector sector (as opposed to the `LVDSstatus` one which refers to specific PMT channels) -- and yes, it should have been `state` instead of `status` but that train has left some time ago;
* an additional flag to state which logic fired the trigger on each cryostat; so far it can be an adder-like value and a majority-like value, or both;
* additional values in the window mode configuration, now supporting the adder triggers in conjunction with the other modes or alone.

These objects have been filled to accommodate ICARUS trigger settings but they may still be suitable for SBND, at least in general terms.
Pulling for review:
* @jzettle: maintainer of the ICARUS trigger board reader and decoding software;
* @mstancar: because I don't have nearly as much interaction with SBND trigger development and I don't know exactly who is the best SBND person for the software part of it (@pgreen135 maybe?); @mstancar should feel free to unload this to somebody else as needed.

This is a necessary prerequisite for the update of ICARUS trigger decoder for Run3, which will be subject of another `icrauscode` PR soon. But I am starting to roll the ball with this one in the meanwhile.

### DIFF
--- a/sbnobj/Common/Trigger/BeamBits.h
+++ b/sbnobj/Common/Trigger/BeamBits.h
@@ -184,7 +184,7 @@ namespace sbn {
     std::string bitName(triggerType bit);
     /// Returns a mnemonic short name for the trigger window mode.
     std::string bitName(triggerWindowMode bit);
-    /// Returns a mnemonic short name for the trigger window mode.
+    /// Returns a mnemonic short name for the gate type.
     std::string bitName(gateSelection bit);
 
     /// @}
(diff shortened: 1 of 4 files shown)


jzettle (approved):
The changes look fine to me.

miquelnebot (commented):
sectorStatus  change not propagated, would need to be modified in icaruscode as well

PR #110, opened 2024-05-21
sbn::ExtraTriggerInfo: added cryostat trigger information

PetrilloAtWork:
ICARUS trigger hardware is going to contribute one additional information (see e.g. [SBNDocDB 36096](https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=36096) or [36264](https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=36264)).

This PR extends the `sbn::ExtraTriggerInfo` data product to accommodate for that new information, that is the time from the opening of the beam gate at which the hardware sees the trigger. This is a more direct quantification of that time than with the White Rabbit timestamp, because once the hardware takes the trigger decision, this time is frozen by the very same hardware, while it takes further time (and jitters) for the decision to reach the timestamping hardware. In the ICARUS hardware the time is measured in FPGA ticks, but the data product should store it in nanoseconds, delegating the decoding of the value upstream.
The protocol is a default value `NoTrigger` that denotes the absence of this information, a class method to test its availability (`hasTrigger()`), and direct access to the data member value (`beamToTrigger`).

The change is backward-compatible in that old data will look like they don't have this part of the trigger information. However, the additional information is mostly diagnostics for trigger study and performance evaluation, and I don't foresee physics analyses directly using it.

Reviewers:
 * @jzettle, maintainer of the ICARUS trigger decoder (which will be subject of PR SBNSoftware/icaruscode#740 to actually _fill_ the information here);
 * @mstancar as point of contact with SBND trigger expert, with the invitation to delegate as appropriate;
 * @dtorretta56 as one of the authors of the firmware change that provides the new information upstream.

### DIFF
--- a/sbnobj/Common/Trigger/ExtraTriggerInfo.cxx
+++ b/sbnobj/Common/Trigger/ExtraTriggerInfo.cxx
@@ -205,58 +205,44 @@ std::ostream& sbn::operator<< (std::ostream& out, ExtraTriggerInfo const& info)
   if (info.triggerLocationBits != 0) {
     out << "\nLocation(s) of trigger:";
     for (std::string const& bitName: names(info.triggerLocation()))
-      out << " " << bitName;
+      out << "  * " << bitName;
   }
-  out << "\nEast cryostat: "
-      << info.cryostats[ExtraTriggerInfo::EastCryostat].triggerCount
-      << " triggers";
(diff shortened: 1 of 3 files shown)


dtorretta56:
OK to proceed in implementing the changes.

jzettle (approved):

PR #113, opened 2024-08-28
CRT Offline Updates

henrylay97:
This PR makes some changes to (SBND-only) CRT reconstruction products motivated by first data studies. It achieves two things:
- Adds space to retain both CRT timing references throughout reconstruction.
- Ensures the types used for timing variables reflects what is needed at that point in reconstruction (namely changing units to ints given that the application of cable delays & referencing to trigger time can make this value negative).

This should be merged together with accompanying PRs in `sbncode` (SBNSoftware/sbncode#471) and `sbndcode` for compilation reasons.


absolution1 (approved):
Looks good.  Thanks @henrylay97

PR #114, opened 2024-11-11
add function to assign CRT-PMT Match classificaiton to sbnobj

aheggest:
This pull request adds a helper function `assignFlashClassification` that assigns a CRT-PMT Match flashClassification using the number of CRT Hits entering or exiting the detector based on the CRT-PMT Time difference value. The function reads in the number of top, side, or bottom CRT hits entering or exiting the detector and returns a [matchType](https://github.com/SBNSoftware/sbnobj/blob/develop/sbnobj/Common/CRT/CRTPMTMatching.hh#L19). This function will be called in [sbncode PR #482](https://github.com/SBNSoftware/sbncode/pull/482) and can replace these lines in the [CRTPMTMatchingUtils.cxx](https://github.com/SBNSoftware/icaruscode/blob/develop/icaruscode/CRT/CRTUtils/CRTPMTMatchingUtils.cxx#L90C1-L117C35) in icaruscode as well. 

Note: The bottom CRT is currently not used in the CRT-PMT Matching for ICARUS, so their counter variables `bottomex` and `bottomen` are left as 0.

### DIFF
--- a/sbnobj/Common/CRT/CRTPMTMatching.hh
+++ b/sbnobj/Common/CRT/CRTPMTMatching.hh
@@ -33,6 +33,42 @@ namespace sbn::crt {
       exSide_enBottom   = 14, ///< Matched with one Bottom CRT hit before the optical flash and matched with one Side CRT hit after the optical flash.
       others            = 9  ///< All the other cases.
       };
+  MatchType assignFlashClassification(uint topen, uint topex, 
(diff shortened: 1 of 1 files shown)


francescopoppi (approved):
All good on my side.
It is good to keep the bottomCRT included, although I am 90% sure we will never implement it due to its timing resolution.

PR #115, opened 2024-11-12
add function to assign CRT-PMT Match classificaiton to sbnobj [release/SBN2024A]

aheggest:
This is a copy of [sbnobj PR#114](https://github.com/SBNSoftware/sbnobj/pull/114) for the production branch.

### DIFF
--- a/sbnobj/Common/CRT/CRTPMTMatching.hh
+++ b/sbnobj/Common/CRT/CRTPMTMatching.hh
@@ -5,8 +5,8 @@
  * @date   June 1 2023.
  */
 
-#ifndef CRTPMTMATCHING_hh_
-#define CRTPMTMATCHING_hh_
+#ifndef SBNOBJ_COMMON_CRTPMTMATCHING_HH
+#define SBNOBJ_COMMON_CRTPMTMATCHING_HH
 
 // C++ includes
 #include <vector>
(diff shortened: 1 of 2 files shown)


sbnobj/Common/CRT/CRTPMTMatching.hh
```diff
@@ -33,6 +33,42 @@ namespace sbn::crt {
       exSide_enBottom   = 14, ///< Matched with one Bottom CRT hit before the optical flash and matched with one Side CRT hit after the optical flash.
       others            = 9  ///< All the other cases.
       };
+  MatchType assignFlashClassification(uint topen, uint topex, 
+				      uint sideen, uint sideex, 
+				      uint bottomen, uint bottomex)
```
PetrilloAtWork, line 38: Some documentation of the function may be useful. Example:
```suggestion
  /**
   * @brief Returns the classification of a TPC object based on the CRT hits associated to it.
   * @param topen number of hits on the top CRT earlier than the object (entering)
   * @param topex number of hits on the top CRT later than the object (exiting)
   * @param sideen number of hits on any side CRT earlier than the object (entering)
   * @param sideex number of hits on any side CRT later than the object (exiting)
   * @param bottomen number of hits on the bottom CRT earlier than the object (entering)
   * @param bottomex number of hits on the bottom side CRT later than the object (exiting)
   * @return a value encoding the classification of the TPC object
   *   
   */
  MatchType assignFlashClassification(uint topen, uint topex, 
				      uint sideen, uint sideex, 
				      uint bottomen, uint bottomex)
```
You may also consider to add assumptions, like if having both entering and exiting  hits from the same side is expected/allowed and how it is considered (I suppose that's different from bottom, side and "top which is also a bit side"), if more than one hit are allowed per side, etc.

PetrilloAtWork (changes requested):
The definition of the function in a header will cause trouble: that needs to be fixed.

I invite you to consider also to:
 * document the function
 * replace the existing header guard with a "standard" one:
   ```cpp
   #ifndef SBNOBJ_COMMON_CRTPMTMATCHING_HH
   #define SBNOBJ_COMMON_CRTPMTMATCHING_HH
   ```
   Preprocessor macro names can easily conflict, so it's better to err on the safe side.

sbnobj/Common/CRT/CRTPMTMatching.hh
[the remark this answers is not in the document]
aheggest: thank you Gianuca, i addressed this comment in https://github.com/SBNSoftware/sbnobj/pull/115/commits/a97b2773d58f57dcb8dbc311613c4a227ed1c361

aheggest: i added some documentation in this commit https://github.com/SBNSoftware/sbnobj/pull/115/commits/a97b2773d58f57dcb8dbc311613c4a227ed1c361

PetrilloAtWork (approved):
Approved!

PR #116, opened 2024-12-03
CRTPMTMatching fix: move function definition 

aheggest:
This PR addresses some final comments by @PetrilloAtWork on sbnobj PR #114. Here I've moved the assignFlashClassification function definition from `sbnobj/Common/CRT/CRTPMTMatching.hh` to a new file `sbnobj/Common/CRT/CRTPMTMatching.cxx`, and also made sure to use `unsigned int` instead of `uint` as suggested in [this comment](https://github.com/SBNSoftware/sbnobj/pull/115#pullrequestreview-2455615721).

### DIFF
--- a/sbnobj/Common/CRT/CMakeLists.txt
+++ b/sbnobj/Common/CRT/CMakeLists.txt
@@ -1,6 +1,7 @@
 cet_make_library(
   SOURCE
     CRTHit.cc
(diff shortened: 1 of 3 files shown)


PetrilloAtWork (approved):
Approved, thank you.

PR #117, opened 2025-01-21
Introduction of CRTT0TaggingInfo

francescopoppi:
This new class introduces additional information which can be used when studying CRT tagged tracks. This object wants to be complementary to the anab::T0 object.
I chatted a bit with @henrylay97 to have a class which can be shared also with SBND people. They will update this in the future.

### DIFF
--- a/sbnobj/Common/CRT/CRTT0TaggingInfo.hh
+++ b/sbnobj/Common/CRT/CRTT0TaggingInfo.hh
@@ -11,7 +11,17 @@
 #include <limits>
 
 // -----------------------------------------------------------------------------
-namespace sbn::crt { struct CRTT0TaggingInfo; }
+namespace sbn::crt { 
+  /// @brief How was the track fitted when matched with a CRT hit.
+  enum class CRTTaggingTrackFit { 
+    pca = 0, // Track fitted with PCA method
+    startEnd = 1, // Track fitted from Start-End vector
+    kalman = 2, // Track fitted using Kalman Filter
+    others = 9 // Track direction evaluated in other mehods
+  };
+      
+  struct CRTT0TaggingInfo;
+}
 /**
  * @brief Additional information on the matching between CRT and TPC tracks.
  * 
@@ -54,12 +64,18 @@ struct sbn::crt::CRTT0TaggingInfo {
   double crossY = NoDistance;
   double crossZ = NoDistance;
 
+  CRTTaggingTrackFit fitType;
+
   /// Fix Coordinate for the CRT Plane
   // e.g. ICARUS case:
   // For Top CRT region 30 Y coordinate is constant: plane=0
   // For Top CRT region 31/32 and Side CRT 40/41/42/43/44/45 X coordinate is constant: plane=1
   // For Top CRT region 33/34 and Side CRT 46/47 Z coordinate is constant: plane=2
-  int plane=NoPlane;  
+  int plane = NoPlane;
+
+  /// Turth level information of the match.
+  /// Default is false for both data and MC. For MC this information is filled if MC truth is available.
+  bool truthMatch = false;
 
   // --- END ---- Data members -------------------------------------------------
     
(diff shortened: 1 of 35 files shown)


henrylay97:
Hi Francesco, thanks again for producing this object, it is definitely important to have to better bookkeep our matching.

SBND actually has both single hit & track matching. I maybe wasn't brilliantly clear offline, my thought was to avoid bloating the size of these objects unnecessarily was to have two objects with names something like:
- CRTHitT0TaggingInfo 
    - the object you propose here
    - renamed (and perhaps could remove singleHit from the members if you wanted)
    - SBND will also plan to make use of it - all we would need to do is add a couple more comments in the header to describe our use of your enums
- CRTTrackT0TaggingInfo (with SBND producing this object when we're ready to)

This is just a thought - I'm happy to go with a single object if you would rather. My worry is the objects would get a fair bit bigger for _every_ match not just for the track ones.

sbnobj/Common/CRT/CRTT0TaggingInfo.hh
```diff
@@ -0,0 +1,105 @@
...
+  // For Top CRT region 33/34 and Side CRT 46/47 Z coordinate is constant: plane=2
+  int singleHitPlane = NoPlane;
+
+  CRTTaggingTrackFit fitType;
+  CRTTaggingMethod  taggingMethod;
+
+
+  /// Turth level information of the match.
+  /// Default is false for both data and MC. For MC this information is filled if MC truth is available.
+  bool truthMatch = false;
```
PetrilloAtWork, line 95: _art_ data products are immutable, so this information would need to be available at matching time.
Some reasons why truth matching is usually separate for the reconstruction information:
 * it may be executed at a later time
 * it may be executed with different parameters/algorithms
 * doesn't infiltrate simulation code into reconstruction code

This flag does not have much information either: it says some algorithm matched the hit to something, but which something is not given. Overall, I would prefer to have the truth matching information saved in a different way — I'd wage, its own data product or association, depending on how the truth information is available.

sbnobj/Common/CRT/classes_def.xml
```diff
@@ -159,4 +160,23 @@
   <!--<class name="art::Assns<recob::OpFlash,sbn::crt::CRTHit,sbn::crt::CRTPMTMatchingInfo>"/>
  <class name="art::Wrapper<art::Assns<recob::OpFlash,sbn::crt::CRTHit,sbn::crt::CRTPMTMatchingInfo> >"/>
  <class name="art::Wrapper<art::Assns<sbn::crt::CRTHit,recob::OpFlash,sbn::crt::CRTPMTMatchingInfo> >"/>-->
+
+  <!-- CRTT0Tagging Information -->
+<enum name="sbn::crt::CRTTaggingMethod"/>
+<enum name="sbn::crt::CRTTaggingTrackFit"/>
+<class name="sbn::crt::CRTT0TaggingInfo"/>
```
PetrilloAtWork, line 167: This `sbn::crt::CRTT0TaggingInfo` needs a `ClassVersion="10"` to enable ROOT I/O schema evolution (see also other objects here, like `sbn::crt::CRTTrack`).

PetrilloAtWork (changes requested):
I support Henry's suggestion expressed in his review.
In addition, I question the truth matching flag in the object — we can talk about it if you want.
Besides a few suggested changes, `ClassVersion` must be added begore we can proceed (see the specific comment in `classes_def.xml`).

francescopoppi:
@henrylay97  I believe the latest changes should reflect your request.
Also @PetrilloAtWork I followed your suggestion and separated MC truth into a different Info object.

henrylay97:
Thanks for the update Francesco - I'm happy with this!

henrylay97 (approved):

kjplows:
Hi @PetrilloAtWork, could you please take a look at this PR? I believe @francescopoppi has addressed your requested changes and I'll test this now but it'd be great if you could check. Many thanks!

kjplows:
trigger build

kjplows:
trigger build LArSoft/lar*@LARSOFT_SUITE_v10_04_07

kjplows:
@PetrilloAtWork is this ok to merge or does the PR need additional work? Cheers!

francescopoppi:
I actually put the comment in the previous comment... This is old, from the larsoft v9 era

PR #118, opened 2025-01-22
Storage of job execution environment in output files [1/4]

PetrilloAtWork:
This is part of the pull request SBNSoftware/sbncode#497.
All discussion goes there.


kjplows:
Hi @PetrilloAtWork , there are conflicts now - could you please check the change in dep from `lardataalg` to `lardataobj` is OK? Thanks!
Pinging @absolution1 and @SFBayLaser here too for review.

PetrilloAtWork:
The change is the other way, from `lardataobj` to `lardataalg`.
This is not a new requirement from this PR, but rather it's a dependency that was overlooked in the past (probably still my fault, but it's a me of several years ago so I repudiate him).

I can rebase to the latest `develop`, with the hope that it won't be needed again later (hint: it likely will).

PR #126, opened 2025-03-27
Ints -> Doubles for low-level CRT reco

henrylay97:
As is described in SBNSoftware/sbndcode#705 we require the low-level CRT objects to store timings as doubles not integers from now on. This is due to calibrated timing delays that are fractional.


kjplows (approved):
Looks good to me, thanks Henry!

kjplows:
trigger build

PR #132, opened 2025-06-20
Feature/hhausner new bnb quality devices

hausnerh:
These changes allow us to implement BNB quality cuts with the data scraped from ACNET


jzennamo (approved):
Just to double check are there any other variables that are in IFBeam that we might care to know about?

hausnerh:
> Just to double check are there any other variables that are in IFBeam that we might care to know about?

No other variables that I know of. These are just the ones that I was informed were useful for the BNB quality figure of merit

kjplows:
Updating `classes_def.xml` (similar problem to sbnanaobj/#142)

PR #135, opened 2025-08-05
Feature/maxd bnb qual cut

maxdubnowski:
Adds additional information to BNBSpillInfo in order to create SBND Figure of Merit for BNB.

### DIFF
--- a/sbnobj/Common/POTAccounting/BNBSpillInfo.h
+++ b/sbnobj/Common/POTAccounting/BNBSpillInfo.h
@@ -13,10 +13,10 @@ namespace sbn {
   class BNBSpillInfo {
   public:
 
-    unsigned int event;
-
     unsigned long int spill_time_s; //!< The IFDB Beam Spill Time, unit sec
     unsigned long int spill_time_ns; //!< The IFDB Beam Spill Time, unit nsec
+
+    unsigned int event;
 
     float FOM; ///! Figure of Merit for BNB
 
--- a/sbnobj/Common/POTAccounting/classes_def.xml
+++ b/sbnobj/Common/POTAccounting/classes_def.xml
@@ -1,6 +1,6 @@
 <lcgdict>
   <class name="sbn::BNBSpillInfo" ClassVersion="15">
-   <version ClassVersion="15" checksum="2707745502"/>
+   <version ClassVersion="15" checksum="28411542"/>
    <version ClassVersion="14" checksum="1668136717"/>
    <version ClassVersion="13" checksum="3323781823"/>
    <version ClassVersion="12" checksum="3285973467"/>


maxdubnowski:
@PetrilloAtWork  Thank you for the comprehensive review of all my code! I see in the Checksum there are 4 values, but I am not sure which ones to get rid of.

PetrilloAtWork:
Your last commit is correct.
In `git diff` you can see that there were three versions/checksums before your changes, and two after.
The three old versions must be left as they are: they are the guarantee that old data can be still read. The newest version (`15`) reflects your latest version, and needs to be retained for the future. The remaining new one is bound to a transient version and should be removed (then the latest version becomes `14` rather than `15`).

PetrilloAtWork:
Sorry to have you go through this additional pain... it seems that while you were making your changes, Joseph too was, and he got to `develop` first.
So there are GIT conflicts to resolve, and you can't just do that in GitHub.
GIT shows two conflicts:
 1. the class `sbn::BNBSpillInfo`: this is a false conflict, it's just that you and Joseph happened to add independent code in the same place. Keep all the lines of code (removing, of course, the heading lines from `diff`), save and be happy.
 2. the `classes_def.xml`: here things are more complicate. Joseph's checksum must be kept (as version `14`), but it's not the final one — in fact, we don't know which the final one is. So the procedure is:
     1. remove your checksum, i.e. pick the file as in `develop` (e.g. `git checkout develop -- sbnobj/Common/POTAccounting/classes_def.xml`)
     2. rebuild your area; GCCXML should notice that `BNBSpillInfo` has changed, update the checksum for you in `classes_def.xml`, and ask you to recompile (it will work like an error)
     3. rebuild once again, and make sure that there are no errors this time
     4. add, commit and push the changed files into your branch: you should now see that the only change in `classes_def.xml` compared to `develop` is a new version, `15`, for `BNBSpillInfo`.

After pushing, GitHub should not report any conflict any more.

I deliberately skipped on the steps to make GIT _see_ the conflict. That may be the real complication, and it's hard to write generic instructions on that.
In general, the procedure is to rebase all your branches to the last release or to `develop` (and hope that `develop` is currently consistent among repositories and working), then clean and rebuild your area. In the rebase step for `sbnobj`, you will see GIT report the conflicts.
You can ask via instant messaging if you incur into showstoppers.

Congratulations: this kind of conflicts is very, very rare :sweat:

sbnobj/Common/POTAccounting/BNBSpillInfo.h
```diff
@@ -6,30 +6,30 @@
...
+
     unsigned long int spill_time_s; //!< The IFDB Beam Spill Time, unit sec
     unsigned long int spill_time_ns; //!< The IFDB Beam Spill Time, unit nsec
 
-    unsigned int event;
-        
-    //BeamQuaility 
-    //quality//BNB
-    
+    float FOM; // Figure of Merit for BNB
```
PetrilloAtWork, line 21: Please make this a Doxygen documentation comment:
```suggestion
    float FOM; ///! Figure of Merit for BNB
```

sbnobj/Common/POTAccounting/BNBSpillInfo.h
```diff
@@ -6,30 +6,30 @@
...
 #include <vector>
 
 namespace sbn {
+
   class BNBSpillInfo {
   public:
-    //typedef enum triggertype { kBNB, kEXT, kUnbiased, kNUMI } TriggerType; //!< Which beamline are we looking at
-    
+
+    unsigned int event;
```
PetrilloAtWork, line 16: This variable was moved from after `spill_time_ns` to here. This changes the layout of the class on disk.
While I think ROOT should be able to cope with this, it's better to keep things simple for it. Please move the `event` data member back to where it was (I guess now just before `FOM`).
Also note that this may change the checksum, so you may need to again perform the routine of manually update `classes_def.xml` to include only one new class version.

PetrilloAtWork (changes requested):
Two more things to change...

PetrilloAtWork (approved):
Good to go. Thank you!

PR #149, opened 2025-10-06
Change tagger storage from set to vector

henrylay97:
CRT Tracks have a `Points` and `Taggers` field. They both describe the 2 or 3 `CRTSpacePoint`s used to create them. The former describes the cartesian xyz of each point, the latter just enumerates which walls those points were in.

The `Taggers` field was originally constructed as a `set` which means the correspondence between `Points` and `Taggers` was lost. This PR amends this to make it a `vector` thus preserving the order.


aantonakis (approved):
Everything looks good to me

kjplows:
trigger build LArSoft/lar*@LARSOFT_SUITE_v10_11_01

kjplows:
trigger build LArSoft/lar*@LARSOFT_SUITE_v10_11_01 SBNSoftware/sbndcode#840

PR #152, opened 2025-10-27
add initilialization for parent member of TrueParticle

mt82:
Added a initialization value to the parent `member` of `TrueParticle`

### DIFF
--- a/sbnobj/Common/Calibration/classes_def.xml
+++ b/sbnobj/Common/Calibration/classes_def.xml
@@ -14,7 +14,8 @@
    <version ClassVersion="11" checksum="903476586"/>
    <version ClassVersion="10" checksum="118230262"/>
   </class>
-  <class name="sbn::TrackInfo" ClassVersion="17">
+  <class name="sbn::TrackInfo" ClassVersion="18">
+   <version ClassVersion="18" checksum="1112650559"/>
    <version ClassVersion="17" checksum="2056578499"/>
    <version ClassVersion="16" checksum="2134249071"/>
    <version ClassVersion="15" checksum="1667792945"/>


PetrilloAtWork (changes requested):
You know what? This type of mistakes is being hunting us since forever, and I am electing you as the victim who's going to pay for it.

Please change the header code removing the constructors and placing all the initialisation in the line the data member is declared.
For example:
```cpp
struct WireInfo {
    uint16_t wire = uint16_t(-1); //!< Wire number
    uint16_t plane = uint16_t(-1); //!< Plane number
    uint16_t tpc = uint16_t(-1); //!< TPC number
    uint16_t channel = uint16_t(-1); //!< Channel number
    int16_t tdc0 = -1; //!< TDC tick of the first ADC value
    std::vector<short> adcs; //!< List of ADC values
  };
```
And guess what? in doing this example I already caught two mistakes (the construction value for `tdc0` was cast to `uint16_t` even if the variable is signed, `int16_t`; and `channel` _was not initialised_).
This is not new rule: [`[CF-155]`](https://sbnsoftware.github.io/sbn/codingconv/CodingConventionsExplained.html#CF155). The idea is that when a person adds a variable in the list they're more likely to notice that all other are being initialised, and do the same for the new one.

If you don't feel like doing that, let me know and I'll add the commit to this branch.

sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
```diff
@@ -190,163 +112,97 @@ namespace sbn {
...
+    float hit_min_time_p1_tpcE = float(-100000.f); //!< Min hit time of track on plane 1 TPC E
+    float hit_max_time_p1_tpcE = float(-100000.f); //!< Max hit time of track on plane 1 TPC E
+    float hit_min_time_p2_tpcE = float(-100000.f); //!< Min hit time of track on plane 2 TPC E
+    float hit_max_time_p2_tpcE = float(-100000.f); //!< Max hit time of track on plane 2 TPC E
+    float hit_min_time_p0_tpcW = float(-100000.f); //!< Min hit time of track on plane 0 TPC W
+    float hit_max_time_p0_tpcW = float(-100000.f); //!< Max hit time of track on plane 0 TPC W
+    float hit_min_time_p1_tpcW = float(-100000.f); //!< Min hit time of track on plane 1 TPC W
+    float hit_max_time_p1_tpcW = float(-100000.f); //!< Max hit time of track on plane 1 TPC W
+    float hit_min_time_p2_tpcW = float(-100000.f); //!< Min hit time of track on plane 2 TPC W
+    float hit_max_time_p2_tpcW = float(-100000.f); //!< Max hit time of track on plane 2 TPC W
```
PetrilloAtWork: You don't need any conversion here: `-100000.f` is explicitly already a `float`.
```suggestion
    float hit_min_time_p0_tpcE = -100000.f; //!< Min hit time of track on plane 0 TPC E
    float hit_max_time_p0_tpcE = -100000.f; //!< Max hit time of track on plane 0 TPC E
    float hit_min_time_p1_tpcE = -100000.f; //!< Min hit time of track on plane 1 TPC E
    float hit_max_time_p1_tpcE = -100000.f; //!< Max hit time of track on plane 1 TPC E
    float hit_min_time_p2_tpcE = -100000.f; //!< Min hit time of track on plane 2 TPC E
    float hit_max_time_p2_tpcE = -100000.f; //!< Max hit time of track on plane 2 TPC E
    float hit_min_time_p0_tpcW = -100000.f; //!< Min hit time of track on plane 0 TPC W
    float hit_max_time_p0_tpcW = -100000.f; //!< Max hit time of track on plane 0 TPC W
    float hit_min_time_p1_tpcW = -100000.f; //!< Min hit time of track on plane 1 TPC W
    float hit_max_time_p1_tpcW = -100000.f; //!< Max hit time of track on plane 1 TPC W
    float hit_min_time_p2_tpcW = -100000.f; //!< Min hit time of track on plane 2 TPC W
    float hit_max_time_p2_tpcW = -100000.f; //!< Max hit time of track on plane 2 TPC W
```

sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
```diff
@@ -1,187 +1,109 @@
...
-    uint16_t tpc; //!< TPC number of hit
-    uint16_t mult; //!< Multiplicity of hit
-    int16_t start; //!< Start tick of hit [ticks]
-    int16_t end; //!< End tick of hit [ticks]
-    bool hasSP; //!< Whether the hit has a SpacePoint
+    float integral = std::numeric_limits<float>::signaling_NaN();    //!< Integral of gaussian fit to ADC values in hit [ADC]
+    float sumadc   = std::numeric_limits<float>::signaling_NaN();    //!< "SummedADC" -- sum of ADC values under gaussian fit [ADC]
+    float width    = std::numeric_limits<float>::signaling_NaN();    //!< Width of fitted gaussian hit [ticks]
+    float goodness = std::numeric_limits<float>::signaling_NaN();    //!< Goodness-of-fit of fitted gaussian hit
+    float time     = std::numeric_limits<float>::signaling_NaN();    //!< Peak time of hit [ticks]
```
PetrilloAtWork: Note that changing the default values of these data members from `-1` to `NaN` is a **breaking change** (for example, it will change the outcome of the test `if (integral >= 0)`).
Also note, however, that `-1` is not a fool-proof default value for `integral`, `sumadc` and `time` (for this last there is kind of a convention that it will be always positive for any meaningful detector activity, but it should not be relied on).

sbnobj/Common/Calibration/TrackCaloSkimmerObj.h
```diff
@@ -1,187 +1,109 @@
...
-      dqdx(-1),
-      rr(-1),
-      i_snippet((uint16_t)-1),
-      ontraj(false),
-      oncalo(false) {}
-
+    HitInfo h;                                                         //!< Hit information by itself
+    float pitch = std::numeric_limits<float>::signaling_NaN();         //!< Pitch of track across wire the hit is on [cm]
+    float dqdx  = std::numeric_limits<float>::signaling_NaN();         //!< Initial computed dq/dx of hit [ADC/cm]
+    float rr    = std::numeric_limits<float>::signaling_NaN();         //!< Residual range of hit along track [cm]
```
PetrilloAtWork: Same **breaking change** here too (but in this case all the quantities _are_ defined positive... except maybe for fluctuations on dQ/dx).

kjplows:
trigger build LArSoft/lar*@LARSOFT_SUITE_v10_14_00

kjplows:
ClassVersion needs updating, doing this now.

PR #158, opened 2025-12-10
Light Calorimetry: Add obj for light calo objects

lynnt20:
Accompanying PRs: SBNSoftware/sbndcode#878, SBNSoftware/sbncode#619, SBNSoftware/sbnanaobj#181


sbnobj/Common/Reco/classes_def.xml
```diff
@@ -78,6 +78,22 @@
   <class name="art::Assns<sbn::crt::CRTHit, recob::Track, void>" />
   <class name="art::Wrapper<art::Assns<sbn::crt::CRTHit, recob::Track, void>>" />
 
+  <class name="sbn::LightCalo">
+  </class>
```
PetrilloAtWork: This needs a checksum and a class version, so that when we change it ROOT knows how to read the old versions. Start with:

```suggestion
  <class name="sbn::LightCalo" classVersion="10" />
```
then build: it will tell you to build again. Rebuild... at this point you should see a fully successful build. Lastly, commit the new `classes_def.xml` to the branch.

sbnobj/Common/Reco/LightCalo.h
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     * Default constructor.
+     */
+    LightCalo() = default;
+    LightCalo(double charge, double light, double energy, int bestplane)
+      : charge(charge)
+      , light(light)
+      , energy(energy)
+      , bestplane(bestplane)
+      {}
```
PetrilloAtWork: You have introduced a value constructor, and I want to make sure you are aware of the implications. I will compare yours with the same class in ["plain old data"](https://en.cppreference.com/w/cpp/language/classes.html#POD_class) version, _with no user-defined constructor_ (yours has two: the value constructor and the default one).

With your version, it's harder for the user to initialise the object only partially (e.g. initialise `charge` but not `light`). This is in general good, but note that partial initialization may still happen (`LightCalo calo; calo.charge = 40000.0;`).
With the POD version, aggregate initialisation can be used (`LightCalo lc { 40000.0, 30000.0, 3.5, 2 };`) that can omit the last arguments. This may or may not be desired. On the other end, POD have such guarantees (can be copied with a `memcpy()`, for example) that allow libraries writing them to disk to take shortcuts. I don't know whether ROOT takes advantage of any of the shortcuts.

In general I recommend POD classes, but that opens for partial initialisation.
So it's up to you and how you feel about it.

PetrilloAtWork (changes requested):
The class version business must be fixed.
I have left also a technical C++ comment: fell completely free to decide how to approach it.

PetrilloAtWork (approved):
I have left two completely optional comments, but the PR is approved.

PR #167, opened 2026-02-16
CRT Blobs

henrylay97:
I am starting to put together slides and PRs to preserve work of mine that lives offline before I leave.

The CRTBlob reconstruction aims to produce objects that represent the totality of all activity happening within a fcl-configurable coincidence window.  They are similar to CRTTracks but without geometric track constraints, and as such can capture all activity from a single source, for example when multiple particles from the same air shower are detected or secondary particles are produced from the muon scattering.

It was used to produce a plot for the detector paper and is therefore worth preserving. It is not run by default in production fcls but does provide fcls for running it independently in the CRT-only workflow or standalone. Nothing prevents it being added to production fcls in future if it is desired for any analyses.

It is documented in slides: https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=45658
Accompanying PR: https://github.com/SBNSoftware/sbndcode/pull/916

### DIFF
--- a/sbnobj/SBND/CRT/CMakeLists.txt
+++ b/sbnobj/SBND/CRT/CMakeLists.txt
@@ -7,6 +7,7 @@ cet_make_library(
     CRTStripHit.cxx
     CRTTrack.cxx
     CRTVeto.cxx
+    CRTBlob.cxx
     FEBData.cxx
     FEBTruthInfo.cxx
   LIBRARIES
--- a/sbnobj/SBND/CRT/CRTBlob.cxx
+++ b/sbnobj/SBND/CRT/CRTBlob.cxx
@@ -13,7 +13,7 @@ namespace sbnd {
       , fTs1        (0.)
       , fTs1Err     (0.)
       , fPE         (0.)
-      , fTaggerHits ({})
+      , fTaggerSPs  ({})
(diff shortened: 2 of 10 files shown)


aantonakis (approved):
Hi Henry, the class looks good and I approve merging the changes. It just needs to be updated to the latest branch. I left a comment about the default values but I think they are sufficient for analysis.

henrylay97:
Great comment from @aantonakis . I noticed it was a theme across SBND CRT objects. Whilst there is no scenario in the producer modules that should ever use the default constructor it still seemed worth changing the default values to more obviously non-physical values across the board.

kjplows:
trigger build LArSoft/lar*@LARSOFT_SUITE_v10_14_02_02 SBNSoftware/sbndaq-artdaq-core@v1_10_06 SBNSoftware/sbndcode#916 SBNSoftware/sbn*@SBN_SUITE_v10_14_02_03

hgreenlee:
trigger build LArSoft/lar*@LARSOFT_SUITE_v10_15_00 SBNSoftware/sbndaq-artdaq-core@v1_10_06 SBNSoftware/sbn*@SBN_SUITE_v10_15_00

PR #170, opened 2026-03-04
SBND Timing Reconstruction Refactor

VCLanNguyen:
## Description 

- Timing reconstruction is refactored in this PR, affecting the decode/reconstruction workflow of CRT/PMT/XA.
- After the refactoring, the @FrameShiftModule@ runs first in the decode chain, outputs timing products to be used at PMT/XA decoder and CRTStrip reconstruction.
- Relevant PMT reconstruction modules at Reco1/Reco2 are updated.
- The timing correction applied at CAFMakeer is undone. This should ressolve this issue: https://github.com/SBNSoftware/sbncode/issues/567
- Details and validation plots can before in the linked docdb.

### Link(s) to docdb describing changes (optional)

https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=46654

### Relevant PR links (optional)

This PR needs the XA decoder PR in sbndcode to go in first:
https://github.com/SBNSoftware/sbndcode/pull/847 

This PR needs to be merged together with this group of PRs:
- sbndcode: https://github.com/SBNSoftware/sbndcode/pull/915
- sbncode: https://github.com/SBNSoftware/sbncode/pull/637
- sbnobj: https://github.com/SBNSoftware/sbnobj/pull/170
- sbnanaobj:https://github.com/SBNSoftware/sbnanaobj/pull/188


VCLanNguyen:
@JosiePaton @PetrilloAtWork can you guys review for CAF?

sbnobj/SBND/Timing/FrameShiftInfo.hh
```diff
@@ -12,32 +12,84 @@
...
+   * 1 - PTB HLT
+   * 2 - No frame found 
+   * std::numeric_limits<uint16_t>::max() - Not initialized            
+   *
+   * Timing Channel incicates which channel the frame is derived from:
+   * SPEC TDC: indicates which input channel 0 to 4
+   * PTB: indicates the High Level Trigger (HLT) type
+   *
    * For more information, see
+   * [SBN DOCDB ???] ?????
```
PetrilloAtWork: I suppose:
```suggestion
   * [SBN DOCDB 45982](https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=45982)
```

sbnobj/SBND/Timing/FrameShiftInfo.hh
```diff
@@ -49,36 +101,64 @@ namespace sbnd::timing {
...
+    uint16_t TimingTypeBeamGate() const { return fTimingTypeBeamGate; }
+    uint16_t TimingChannelBeamGate() const { return fTimingChannelBeamGate; }
+
+    uint64_t FrameEtrig() const { return fFrameEtrig; }
+    uint16_t TimingTypeEtrig() const { return fTimingTypeEtrig; }
+    uint16_t TimingChannelEtrig() const { return fTimingChannelEtrig; }
+
+    uint64_t FrameDefault() const { return fFrameDefault; }
+    uint16_t TimingTypeDefault() const { return fTimingTypeDefault; }
+    uint16_t TimingChannelDefault() const { return fTimingChannelDefault; }
```
PetrilloAtWork, line 151: A note, maybe just for the future: this class has private data accessible only via member functions, which means that users will have to interact with the latter; but only the former are documented inline, which makes one additional step (involving parsing source code) to get to the documentation.
I would suggest the interface to be documented with priority over the internal implementation (data in this case); ideally, both, although it adds maintenance burden.

sbnobj/SBND/Timing/TimingInfo.hh
```diff
@@ -10,9 +10,13 @@
 #define SBND_TIMINGINFO_HH
 
 #include <stdint.h>
+#include <limits>
 
 namespace sbnd::timing {
 
+  static constexpr uint64_t kSecondInNanoseconds = static_cast<uint64_t>(1e9);           ///< 1s = 1,000,000,000ns
```
PetrilloAtWork: I suggest:
```suggestion
  static constexpr uint64_t kSecondInNanoseconds = 1'000'000'000;           ///< 1s = 1,000,000,000ns
```
(or `1'000'000'000ULL` if we want full pedantry).

PetrilloAtWork (commented):
This is a very strong breaking change, which can be focussed into two places.
1. `sbnd::timing::FrameShiftInfo`: I think it's fair to say it's a completely different class, at the point where one should think what happens when this new class is initialised with old data. ROOT will convert what it can, and from the documentation it seems the conversion will be wrong (turning from a shift — `double` — to a timestamp — `uint64_t`). I _strongly recommend_ you to think about this issue, and act accordingly. Two possibilities that come to my mind are:
    1. Remove version 10 from `classes_def.xml`; I am not sure which type of failure this will trigger when reading old data, but I believe there will be some. This would need to be verified.
    2. Write explicit conversion code ("ROOT I/O rule") handling the conversion. In the minimal case, this rule would throw an exception claiming the "upgrade" is not supported.
2. `sbnd::timing::TimingInfo`: undefined values change from `0` to something like `INT_MAX`. Again old data will be interpreted differently. An I/O rule changing the data members that are `0` to the new undefined value should be arranged, and the code should be surveyed to make sure that the magic undefined value `0` is not implicitly used.

VCLanNguyen:
> This is a very strong breaking change, which can be focussed into two places.
> 1. `sbnd::timing::FrameShiftInfo`: I think it's fair to say it's a completely different class, at the point where one should think what happens when this new class is initialised with old data. ROOT will convert what it can, and from the documentation it seems the conversion will be wrong (turning from a shift — `double` — to a timestamp — `uint64_t`). 

I agree that this is a break changing for SBND workflow. We refactored the entire decoding on raw data to unify the timing reconstruction. Understandably, this is a clean up attempt of bad code writing in the first place.
So  initialising this new class with 'old decoded data' should break.

However, I don't fully understand all of your suggestions below, so adding my question to each your point below.

>     1. Remove version 10 from `classes_def.xml`; I am not sure which type of failure this will trigger when reading old data, but I believe there will be some. This would need to be verified.

Is the suggestion to keep or remove this [line](https://github.com/SBNSoftware/sbnobj/pull/170/changes#diff-ce96559a79943bda5eef0043ca4c99b8860f637babd98841e36439f047cf36ddL15)? 

>     2. Write explicit conversion code ("ROOT I/O rule") handling the conversion. In the minimal case, this rule would throw an exception claiming the "upgrade" is not supported.

Agree that this can be a clean suggestion although I do not know what ROOT I/O rules are. 
- Can you point me to an example? 
- Where does the I/O need to be implemented, for example, in the module where the classes are created?


> 2. `sbnd::timing::TimingInfo`: undefined values change from `0` to something like `INT_MAX`. Again old data will be interpreted differently. An I/O rule changing the data members that are `0` to the new undefined value should be arranged, and the code should be surveyed to make sure that the magic undefined value `0` is not implicitly used.
- So in the module, the classes are initialised like [this](https://github.com/SBNSoftware/sbndcode/blob/93faf62c6fe79833669b3b46cd2f8c0f1b427c03/sbndcode/Timing/FrameShift/FrameShift_module.cc#L314-L321) so 0 is not explicitly called anywhere. Is that sufficient?

VCLanNguyen:
After discussion with SBND sim/reco conveners, we understand that this workflow is not backward incompatibility since the change stems all the way from raw data through decode and reco.
This is something we would have to make note to tell analysers "from this release onwards, this code version XX will not be compatible with old data version YY".

PetrilloAtWork:
> from this release onwards, this code version XX will not be compatible with old data version YY

That's what I call "backward incompatibility" :-D
But, wording aside, you have proven SBND is aware of what this involves, that is what I wanted to make sure.

PetrilloAtWork:
> I agree that this is a break changing for SBND workflow. We refactored the entire decoding on raw data to unify the timing reconstruction. Understandably, this is a clean up attempt of bad code writing in the first place. So initialising this new class with 'old decoded data' should break.
[...]
> > 1. Remove version 10 from `classes_def.xml`; I am not sure which type of failure this will trigger when reading old data, but I believe there will be some. This would need to be verified.
> > ```
> 
> Is the suggestion to keep or remove this [line](https://github.com/SBNSoftware/sbnobj/pull/170/changes#diff-ce96559a79943bda5eef0043ca4c99b8860f637babd98841e36439f047cf36ddL15)?

My suggestion was to remove that line, in the hope that ROOT would refuse to load the object. The goal is to make the "breaking" as loud and noticeable as possible, as opposed to silent.

> > ```
> > 2. Write explicit conversion code ("ROOT I/O rule") handling the conversion. In the minimal case, this rule would throw an exception claiming the "upgrade" is not supported.
> > ```
> 
> Agree that this can be a clean suggestion although I do not know what ROOT I/O rules are.
> 
> * Can you point me to an example?
> * Where does the I/O need to be implemented, for example, in the module where the classes are created?

This is an example from [`recob::Hit`](https://github.com/LArSoft/lardataobj/blob/4a807a0dfc7a857a872562dcd1213d6c6e57f618/lardataobj/RecoBase/classes_def.xml#L240-L251) where `fSummedADC` was renamed into `fROISummedADC`. I/O rules make sense only if you can correctly convert the value — probably you can't this time, but you'll tell. You may even just throw an exception in the rule code, I suppose.

All this is in the Kingdom of Up To You.

PetrilloAtWork:
> > 2. `sbnd::timing::TimingInfo`: undefined values change from `0` to something like `INT_MAX`. Again old data will be interpreted differently. An I/O rule changing the data members that are `0` to the new undefined value should be arranged, and the code should be surveyed to make sure that the magic undefined value `0` is not implicitly used.
> 
> * So in the module, the classes are initialised like [this](https://github.com/SBNSoftware/sbndcode/blob/93faf62c6fe79833669b3b46cd2f8c0f1b427c03/sbndcode/Timing/FrameShift/FrameShift_module.cc#L314-L321) so 0 is not explicitly called anywhere. Is that sufficient?

The problem is in user code which _reads_ the new data and might decide whether the data is invalid comparing it to `0` (bad: they should have used `TimingInfo::InvalidTimestamp`!). Since now the invalid value is `max()`, the old code will see that it is not `0` and will consider it a valid timestamp. That might actually be solved with a I/O rule (as above) that in moving from version 10 to 11 goes
```cpp
fTdcCrtt1 = (onfile.fTdcCrtt1 == 0)? sbn::timing::TimingInfo::kInvalidTimestamp: onfile.fTdcCrtt1;
```
Unfortunately you don't have control on user code (although maybe you can safely assume nobody is using that).

VCLanNguyen:
> > > 2. `sbnd::timing::TimingInfo`: undefined values change from `0` to something like `INT_MAX`. Again old data will be interpreted differently. An I/O rule changing the data members that are `0` to the new undefined value should be arranged, and the code should be surveyed to make sure that the magic undefined value `0` is not implicitly used.
> > 
> > 
> > 
> > * So in the module, the classes are initialised like [this](https://github.com/SBNSoftware/sbndcode/blob/93faf62c6fe79833669b3b46cd2f8c0f1b427c03/sbndcode/Timing/FrameShift/FrameShift_module.cc#L314-L321) so 0 is not explicitly called anywhere. Is that sufficient?
> 
> The problem is in user code which _reads_ the new data and might decide whether the data is invalid comparing it to `0` (bad: they should have used `TimingInfo::InvalidTimestamp`!). Since now the invalid value is `max()`, the old code will see that it is not `0` and will consider it a valid timestamp. That might actually be solved with a I/O rule (as above) that in moving from version 10 to 11 goes
> 
> ```c++
> fTdcCrtt1 = (onfile.fTdcCrtt1 == 0)? sbn::timing::TimingInfo::kInvalidTimestamp: onfile.fTdcCrtt1;
> ```
> 
> Unfortunately you don't have control on user code (although maybe you can safely assume nobody is using that).

Updated class_def.xml to have i/o rule for the sbnd::timing::TimingInfo objet

VCLanNguyen:
> > I agree that this is a break changing for SBND workflow. We refactored the entire decoding on raw data to unify the timing reconstruction. Understandably, this is a clean up attempt of bad code writing in the first place. So initialising this new class with 'old decoded data' should break.
> > [...]
> > > 1. Remove version 10 from `classes_def.xml`; I am not sure which type of failure this will trigger when reading old data, but I believe there will be some. This would need to be verified.
> > 
> > 
> > Is the suggestion to keep or remove this [line](https://github.com/SBNSoftware/sbnobj/pull/170/changes#diff-ce96559a79943bda5eef0043ca4c99b8860f637babd98841e36439f047cf36ddL15)?
> 
> My suggestion was to remove that line, in the hope that ROOT would refuse to load the object. The goal is to make the "breaking" as loud and noticeable as possible, as opposed to silent.
> 
> > > ```
> > > 2. Write explicit conversion code ("ROOT I/O rule") handling the conversion. In the minimal case, this rule would throw an exception claiming the "upgrade" is not supported.
> > > ```
> > 
> > 
> > Agree that this can be a clean suggestion although I do not know what ROOT I/O rules are.
> > 
> > * Can you point me to an example?
> > * Where does the I/O need to be implemented, for example, in the module where the classes are created?
> 
> This is an example from [`recob::Hit`](https://github.com/LArSoft/lardataobj/blob/4a807a0dfc7a857a872562dcd1213d6c6e57f618/lardataobj/RecoBase/classes_def.xml#L240-L251) where `fSummedADC` was renamed into `fROISummedADC`. I/O rules make sense only if you can correctly convert the value — probably you can't this time, but you'll tell. You may even just throw an exception in the rule code, I suppose.
> 
> All this is in the Kingdom of Up To You.

So I'm not sure that on SBND we want a break that is that loud.

To clarify, this is only incompatible if users are looking at timing objects, the rest of reco objects (Flash PEs, vertex, etc.) are not impacted. So we'd like to keep the 'new CAF workflow' not to break entirely if users need to use it on 'old data'.

I've discussed with the SBND reco conveners to do documentation on docdb+announcement+release documentation to make sure that the backward-incompatibility only affects timing-related reco objects.

VCLanNguyen:
Confirmed that sbnd::timing::TimingInfo with value 0 is converted correctly to uint64_t max so the io rule works
image

VCLanNguyen:
The io rule was the last thing requested by @PetrilloAtWork for review. Now it's done, the PR can be considered approved.

PetrilloAtWork (approved):
Formal approval too.

